PR #31, opened 2020-05-29
Properly handle wush error messages

### LINKED ISSUE #28: Bad error message for coder sh offline environment
Screen Shot 2020-05-25 at 2 41 39 PM

This is what you get if you coder sh into an environment that was turned off by auto off.

coadler:
Closes #28 

Prints the error returned from wush-lite instead of identifying it as an invalid stream id

Before:
image

After:
image

### DIFF
--- a/wush/client.go
+++ b/wush/client.go
@@ -3,8 +3,9 @@ package wush
 import (
 	"context"
 	"encoding/base64"
-	"fmt"
+	"encoding/json"
 	"io"
+	"io/ioutil"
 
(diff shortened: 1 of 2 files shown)


wush/client.go
```diff
@@ -161,6 +167,23 @@ func NewClient(ctx context.Context, conn *websocket.Conn) *Client {
 	return c
 }
 
+func (c *Client) handleStreamError(body []byte) error {
```
scsmithr, line 170: Does this need to be a method on Client?

PR #38, opened 2020-06-12
Usage metric pushing

coadler:


internal/activity/copy.go
```diff
@@ -0,0 +1,36 @@
+package activity
+
+import "io"
+
+// Copy copyes src to dst, pushing activity when bytes are read from src.
+func Copy(p *Pusher, dst io.Writer, src io.Reader) (written int64, err error) {
```
ammario, line 6: A better way to represent this would be adding a `Writer(w io.Writer) io.Writer` method to pusher. Then, native io.Copy (or any transport) can be used.

cmd/coder/shell.go
```diff
@@ -142,7 +143,12 @@ func runCommand(ctx context.Context, envName string, command string, args []stri
 	go func() {
 		stdin := process.Stdin()
 		defer stdin.Close()
-		io.Copy(stdin, os.Stdin)
+
+		ap := &activity.Pusher{Source: "ssh", EnvID: env.ID, Client: entClient}
+		defer ap.Start()()
+
+		wr := activity.Writer(ap, stdin)
```
ammario, line 150: Why not put the Writer method on the Activity Pusher? We get better autocomplete discoverability and a shorter invokation:

```suggestion
		wr := ap.Writer(stdin)
```

internal/activity/copy.go
```diff
@@ -0,0 +1,17 @@
+package activity
```
ammario: File name is no longer correct

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+const pushInterval = time.Minute
+
+// Pusher pushes activity metrics no more than once per pushInterval. Pushes
+// within the same interval are a no-op.
+type Pusher struct {
+	Source string
+	EnvID  string
+	Client *entclient.Client
+
+	state int64
```
ammario, line 21: I think any time a CAS is used a good justification should be nearby. I see the code is trying to achieve a single flight, and so a mutex is avoided. A more readable alternative would be https://pkg.go.dev/golang.org/x/time/rate?tab=doc.

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func (p *Pusher) Push() {
+	if atomic.CompareAndSwapInt64(&p.state, 0, 1) {
+		err := p.Client.PushActivity(p.Source, p.EnvID)
+		if err != nil {
+			flog.Error("failed to push activity: %s", err.Error())
```
ammario, line 28: `failed to` prefixes are illegal

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+type Pusher struct {
+	Source string
+	EnvID  string
+	Client *entclient.Client
+
+	state int64
+}
+
+func (p *Pusher) Push() {
+	if atomic.CompareAndSwapInt64(&p.state, 0, 1) {
```
ammario, line 25: Control flow inverted

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+			case <-tick.C:
+				atomic.StoreInt64(&p.state, 0)
+			}
+
+		}
+	}()
+
+	return func() {
+		cancel()
+	}
```
ammario, line 54: Just return cancel itself.

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+		err := p.Client.PushActivity(p.Source, p.EnvID)
+		if err != nil {
+			flog.Error("failed to push activity: %s", err.Error())
+		}
+	}
+}
+
+// Start starts the reset routine. It resets the state every pushInterval,
+// allowing Push to push a new activity.
+func (p *Pusher) Start() (end func()) {
```
ammario, line 35: Usually, these are better structured when they take a context. The code ends up with less `WithCancel`s and calls to `cancel()`.

internal/entclient/activity.go
```diff
@@ -0,0 +1,19 @@
+package entclient
+
+import "net/http"
+
+func (c Client) PushActivity(source string, envID string) error {
+	res, err := c.request("POST", "/api/metrics/usage/push", map[string]string{
```
ammario, line 6: We're not consistent about "Activity" or "Metrics"

internal/activity/pusher.go
```diff
@@ -0,0 +1,55 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
```
ammario, line 37: The creation of the goroutine should be as high up the stack as possible

- gives callers better ability to orchestrate (e.g WaitGroup)
- non-synchronous functions are always slightly surprising
- the calling code is always more readable

PR #41, opened 2020-06-12
Fix coder sh freeze

### LINKED ISSUE #40: Coder sh freezes
I just had this happen for the first time ever, ~20 minutes after updating to the wsep version. After returning to my terminal after a bit of idling, it's completely unresponsive. No inputs are registered, ctrl+c / ctrl+d won't escape, I had to close the terminal and reconnect.

The old wush implementation would just kick me back out to my local session periodically, so I assume the disconnect that is happening isn't being gracefully treated with the wsep implementation.

cmoog:
Fixes #40

### DIFF
--- a/cmd/coder/shell.go
+++ b/cmd/coder/shell.go
@@ -99,7 +99,7 @@ func (cmd *shellCmd) Run(fl *pflag.FlagSet) {
 		os.Exit(exitErr.Code)
 	}
 	if err != nil {
-		flog.Fatal("run command: %v Is it online?", err)
+		flog.Fatal("run command: %v. Is %q online?", err, envName)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cmd/coder/shell.go
```diff
@@ -142,10 +145,26 @@ func runCommand(ctx context.Context, envName string, command string, args []stri
...
+		}
 	}()
-	go io.Copy(os.Stdout, process.Stdout())
-	go io.Copy(os.Stderr, process.Stderr())
-
-	return process.Wait()
+	go func() {
+		_, err := io.Copy(os.Stdout, process.Stdout())
+		if err != nil {
+			cancel()
```
ammario, line 156: nitpick: If one of these closes both do... It's fair to just defer cancel() at the top of both

cmoog: I don't think it's good to depend on that behavior though...

PR #49, opened 2020-06-16
Sync fails if path leading to remoteDir doesn't already exist

Russtopia:
I just noticed that setting up sync on a new Environment requires going in first with `coder sh` to create intermediate subdirs of the desired remote path. This patch makes that unnecessary, automatically creating the remote path if not already existent.

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -71,7 +71,7 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
 	return nil
 }
 
-func (s Sync) remoteRm(ctx context.Context, remote string) error {
+func (s Sync) remoteCmd(ctx context.Context, prog string, args ...string) error {
 	conn, err := s.Client.DialWsep(ctx, s.Env)
 	if err != nil {
 		return err
@@ -80,8 +80,8 @@ func (s Sync) remoteRm(ctx context.Context, remote string) error {
 
 	execer := wsep.RemoteExecer(conn)
 	process, err := execer.Start(ctx, wsep.Command{
-		Command: "rm",
-		Args:    []string{"-rf", remote},
+		Command: prog,
+		Args:    args,
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


internal/sync/sync.go
```diff
@@ -67,6 +67,37 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
...
+	err = process.Wait()
+	if code, ok := err.(wsep.ExitError); ok {
+		return fmt.Errorf("mkdir exit status: %v", code)
+	}
+	if err != nil {
+		return xerrors.Errorf("execution failure: %w", err)
+	}
+	return nil
+}
+
```
Russtopia, line 100: Blatant cut 'n paste of remoteRm() just below.

internal/sync/sync.go
```diff
@@ -67,6 +67,37 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
 	return nil
 }
 
+func (s Sync) remoteMkDir() error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
```
scsmithr, line 72: Context should be passed in.

internal/sync/sync.go
```diff
@@ -67,6 +67,37 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
...
+	go io.Copy(os.Stderr, process.Stdout())
+
+	err = process.Wait()
+	if code, ok := err.(wsep.ExitError); ok {
+		return fmt.Errorf("mkdir exit status: %v", code)
+	}
+	if err != nil {
+		return xerrors.Errorf("execution failure: %w", err)
+	}
+	return nil
```
scsmithr, line 98: Would it make sense to move the parts that's shared with `remoteRm` to its own function? Maybe something like `remoteCommand(ctx context.Context, command string, args []string) error`.

Russtopia: Seems like a good candidate to refactor. I'll look at it

PR #50, opened 2020-06-16
Add heartbeat

cmoog:


cmoog:
This still does not fix the primary kick case from stale `envagent` pods.

PR #54, opened 2020-06-19
login: detect missing protocol in URL

### LINKED ISSUE #53: login: Malformed login URL arg breaks login, sh and sync
As a naive user I tried

```coder login localhost:8080```
```coder login master.cdr.dev```

The browser auto-launch of the login URL is malformed in the above cases (and the text on terminal if autolaunch fails)

If I specify the full URL eg.
```coder login http://localhost:8080```

the URL generated is correct, and subsequent ```coder sh``` and ```coder sync``` ops work.

Recommend adding logic to prepend `http://` if not specified by the user.

### PR DESCRIPTION
**Pull request by Russtopia**
Fixes #53 

### What This Does

Detects a user entering a URL without the protocol field, eg. 'localhost:8080' instead of 'http://localhost:8080' which would result in a malformed login URL, and the stored state using incorrect API paths for all coder-cli operations (eg. sync, sh) which give obscure rsync errors.

Since the cli library dispatches sub-commands to the login.Run() function as non-writable array fl.Arg(), the URL can't be patched up silently to be correct, so instead this patch detects the missing protocol prefix in the URL, and prints out a help usage modified to explain that http[s]:// is required.

### DIFF
--- a/cmd/coder/login.go
+++ b/cmd/coder/login.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 
 	"github.com/pkg/browser"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/coder/login.go [hunk 1]
```diff
@@ -22,13 +23,13 @@ type loginCmd struct {
...
-		Usage: "[Coder Enterprise URL]",
+		Usage: "[Coder Enterprise URL eg. http://my.coder.domain/ ]",
 		Desc:  "authenticate this client for future operations",
 	}
 }
 func (cmd loginCmd) Run(fl *pflag.FlagSet) {
 	rawURL := fl.Arg(0)
-	if rawURL == "" {
+	if rawURL == "" || !strings.HasPrefix(rawURL, "http") {
 		exitUsage(fl)
```
Russtopia, line 33: I tested against http://localhost:8080, http://master.cdr.dev/ and https://master.cdr.dev/ and all seem to work (ie., master.cdr.dev auto-downgrades the login URL to http, which I assume is desired).

PR #56, opened 2020-06-24
Update API request, print mult. envs in tabs, handle non 200 response

anthonyshull:


cmd/coder/urls.go
```diff
@@ -0,0 +1,73 @@
...
+	if len(devURLs) == 0 {
+		fmt.Printf("no dev urls were found for environment: %s\n", envName)
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', tabwriter.TabIndent)
+	for _, devURL := range devURLs {
+		fmt.Fprintf(w, "%s\t%s\t%s\n", devURL.URL, devURL.Port, devURL.Access)
+	}
+	w.Flush()
+}
```
Russtopia, line 73: LGTM.

Not needed now, but for future would it be useful to offer a -q (quiet) switch to let users get subsets of the output (say just the url and not port,access)?

Issue #57, opened 2020-06-24
Add create and del subcommands to manage devurls


Russtopia:
Proposal to add new subcommands,
*  `create <env_name> <port> <access>`
* `del <env_name> <port>`

... allowing management of devurls and a full CRUD capability via coder-cli.

* Requires prior login as a user with access to admin <env_name>
* Subject to same site-install/org devurl access policies as doing this through the web frontend
* The `create` subcommand should be an idempotent operation. Calling `create` on an existing `<env_name> <port>` will update the devurl's `<access>` if a new value is specified, otherwise the command will do nothing and return success.

These commands would be useful both for end users and for integration testing with cemanager, enabling scripted devurl testing to ensure the API endpoints haven't broken by changes either in coder-cli or enterprise.

The `create` and `del` subcommands <strike>may require work on cdr/enterprise as well but the existing endpoints used by the web FE can/should be usable.</strike> should be able to use existing endpoints available for the web frontend.

TBD: subcommand names. `create` and `del` may be too generic if coder-cli eventually manages other resources. Will implement as these for now, but suggest `addurl` and `delurl` instead?

PR #60, opened 2020-06-25
Devurl crud ops

Russtopia:
This implements the new subcommands proposed in #57 

### Example Usage

```
$ ./coder urls russ-env1
http://3000-russ-env1-admin.russtopia-rlmfuntoo.dogfood.c8s.io/ 3000 PRIVATE
```
```
$ ./coder urls create russ-env1 3001 --access=org
Adding devurl for port 3001
$ ./coder urls russ-env1
http://3000-russ-env1-admin.russtopia-rlmfuntoo.dogfood.c8s.io/ 3000 PRIVATE
http://3001-russ-env1-admin.russtopia-rlmfuntoo.dogfood.c8s.io/ 3001 ORG
```
```
$ ./coder urls create russ-env1 3002 -a group
Invalid access level
Usage: coder urls create <env name> <port> [--access <level>]

create/update a devurl for external access

coder urls create flags:
      -a, --access string   [private | org | authed | public] set devurl access (default "private")
```
```
$ ./coder urls del russ-env1 8000
2020-06-25 09:38:25 FATAL	No devurl found for port 8000
```
```
$ ./coder urls -h
Usage: coder urls <env name>

get all development urls for external access

Commands:
	create  create/update a devurl for external access
	del     delete a devurl
```
```
$ ./coder urls create -h
Usage: coder urls create <env name> <port> [--access <level>]

create/update a devurl for external access

coder urls create flags:
      -a, --access string   [private | org | authed | public] set devurl access (default "private")
```
```
$ ./coder help urls del
Usage: coder urls del <env name> <port>

delete a devurl
```
```
$ ./coder urls create russ-env1 99999 -a=public
bad port
Usage: coder urls create <env name> <port> [--a access]

create/update a devurl for external access

coder urls create flags:
      -a, --access string   [private | org | authed | public] set devurl access (default "private")
```

### DIFF
--- a/cmd/coder/sync.go
+++ b/cmd/coder/sync.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -30,10 +28,6 @@ func (cmd *syncCmd) Spec() cli.CommandSpec {
 func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 	fl.BoolVarP(&cmd.init, "init", "i", false, "do initial transfer and exit")
 }
-
-// See https://lxadm.com/Rsync_exit_codes#List_of_standard_rsync_exit_codes.
-var IncompatRsync = errors.New("rsync: exit status 2")
-var StreamErrRsync = errors.New("rsync: exit status 12")
 
 func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 	var (
@@ -81,11 +75,7 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 		err = s.Run()
 	}
 
-	if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", IncompatRsync) {
-		flog.Fatal("no compatible rsync present on remote machine")
-	} else if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", StreamErrRsync) {
-		flog.Fatal("error in rsync protocol datastream (no installed remote rsync?)")
-	} else {
-		flog.Fatal("sync: %v", err)
+	if err != nil {
+		flog.Fatal("%v", err)
 	}
 }
--- a/cmd/coder/sync.go
+++ b/cmd/coder/sync.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -31,7 +32,8 @@ func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 }
 
 // See https://lxadm.com/Rsync_exit_codes#List_of_standard_rsync_exit_codes.
-var NoRsync = errors.New("rsync: exit status 2")
+var IncompatRsync = errors.New("rsync: exit status 2")
+var StreamErrRsync = errors.New("rsync: exit status 12")
 
 func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 	var (
@@ -79,8 +81,10 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 		err = s.Run()
 	}
 
-	if err == NoRsync {
+	if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", IncompatRsync) {
 		flog.Fatal("no compatible rsync present on remote machine")
+	} else if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", StreamErrRsync) {
+		flog.Fatal("error in rsync protocol datastream (no installed remote rsync?)")
 	} else {
 		flog.Fatal("sync: %v", err)
 	}
--- a/cmd/coder/urls.go
+++ b/cmd/coder/urls.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/pflag"
@@ -15,10 +17,145 @@ import (
 
 type urlsCmd struct{}
 
+// DevURL is the parsed json response record for a devURL from cemanager
 type DevURL struct {
+	ID     string `json:"id"`
 	URL    string `json:"url"`
 	Port   string `json:"port"`
 	Access string `json:"access"`
+}
+
+var urlAccessLevel = map[string]string{
+	//Remote API endpoint requires these in uppercase
+	"PRIVATE": "Only you can access",
+	"ORG":     "All members of your organization can access",
+	"AUTHED":  "Authenticated users can access",
+	"PUBLIC":  "Anyone on the internet can access this link",
+}
+
+func portIsValid(port string) bool {
+	p, err := strconv.ParseUint(port, 10, 16)
+	if p < 1 {
+		// port 0 means 'any free port', which we don't support
+		err = strconv.ErrRange
+	}
+	if err != nil {
+		fmt.Println("Invalid port")
+	}
+	return err == nil
+}
+
+func accessLevelIsValid(level string) bool {
+	_, ok := urlAccessLevel[level]
+	if !ok {
+		fmt.Println("Invalid access level")
+	}
+	return ok
+}
+
+type createSubCmd struct {
+	access string
+}
+
+func (sub *createSubCmd) RegisterFlags(fl *pflag.FlagSet) {
+	fl.StringVarP(&sub.access, "access", "a", "private", "[private | org | authed | public] set devurl access")
+}
+
+func (sub createSubCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "create",
+		Usage: "<env name> <port> [--access <level>]",
+		Desc:  "create/update a devurl for external access",
+	}
+}
+
+// Run creates or updates a devURL, specified by env ID and port
+// (fl.Arg(0) and fl.Arg(1)), with access level (fl.Arg(2)) on
+// the cemanager.
+func (sub createSubCmd) Run(fl *pflag.FlagSet) {
+	envName := fl.Arg(0)
+	port := fl.Arg(1)
+	access := fl.Arg(2)
+
+	if envName == "" {
+		exitUsage(fl)
+	}
+
+	if !portIsValid(port) {
+		exitUsage(fl)
+	}
+
+	access = strings.ToUpper(sub.access)
+	if !accessLevelIsValid(access) {
+		exitUsage(fl)
+	}
+
+	entClient := requireAuth()
+
+	env := findEnv(entClient, envName)
+
+	_, found := devURLID(port, urlList(envName))
+	if found {
+		fmt.Printf("Updating devurl for port %v\n", port)
+	} else {
+		fmt.Printf("Adding devurl for port %v\n", port)
+	}
+
+	err := entClient.UpsertDevURL(env.ID, port, access)
+	if err != nil {
+		flog.Error("upsert devurl: %s", err.Error())
+	}
+}
+
+type delSubCmd struct{}
+
+func (sub delSubCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "del",
+		Usage: "<env name> <port>",
+		Desc:  "delete a devurl",
+	}
+}
+
+// devURLID returns the ID of a devURL, given the env name and port.
+// ("", false) is returned if no match is found.
+func devURLID(port string, urls []DevURL) (string, bool) {
+	for _, url := range urls {
+		if url.Port == port {
+			return url.ID, true
+		}
+	}
+	return "", false
+}
+
+// Run deletes a devURL, specified by env ID and port, from the cemanager.
+func (sub delSubCmd) Run(fl *pflag.FlagSet) {
+	envName := fl.Arg(0)
+	port := fl.Arg(1)
+
+	if envName == "" {
+		exitUsage(fl)
+	}
+
+	if !portIsValid(port) {
+		exitUsage(fl)
+	}
+
+	entClient := requireAuth()
+
+	env := findEnv(entClient, envName)
+
+	urlID, found := devURLID(port, urlList(envName))
+	if found {
+		fmt.Printf("Deleting devurl for port %v\n", port)
+	} else {
+		flog.Fatal("No devurl found for port %v", port)
+	}
+
+	err := entClient.DelDevURL(env.ID, urlID)
+	if err != nil {
+		flog.Error("delete devurl: %s", err.Error())
+	}
 }
 
 func (cmd urlsCmd) Spec() cli.CommandSpec {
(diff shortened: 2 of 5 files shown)


cmd/coder/sync.go
```diff
@@ -31,7 +32,8 @@ func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 }
 
 // See https://lxadm.com/Rsync_exit_codes#List_of_standard_rsync_exit_codes.
-var NoRsync = errors.New("rsync: exit status 2")
+var errIncompatRsync = errors.New("rsync: exit status 2")
+var errStreamErrRsync = errors.New("rsync: exit status 12")
```
Russtopia, line 36: One of the go linters complained these weren't named 'correctly'. Can change back to IncompatRsync, StreamErrRsync if we don't agree.

cmd/coder/sync.go
```diff
@@ -79,8 +81,10 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 		err = s.Run()
 	}
 
-	if err == NoRsync {
+	if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", errIncompatRsync) {
 		flog.Fatal("no compatible rsync present on remote machine")
+	} else if fmt.Sprintf("%v", err) == fmt.Sprintf("%v", errStreamErrRsync) {
+		flog.Fatal("error in rsync protocol datastream (no installed remote rsync?)")
```
Russtopia, line 87: These changes were for error checking fixed in another branch. Not strictly to do with this review

cmd/coder/urls.go
```diff
@@ -14,13 +14,120 @@ import (
...
 	URL    string `json:"url"`
 	Port   string `json:"port"`
 	Access string `json:"access"`
 }
 
+var urlAccessLevel = map[string]string{
+	"PRIVATE": "Only you can access",
+	"ORG":     "All members of your organization can access",
+	"AUTHED":  "Authenticated users can access",
+	"PUBLIC":  "Anyone on the internet can access this link",
```
Russtopia, line 32: The values don't technically matter currently. Validation of access levels just checks user actions against the keys.

cmd/coder/urls.go
```diff
@@ -65,6 +174,14 @@ func (cmd urlsCmd) Run(fl *pflag.FlagSet) {
...
 	}
 
+	return devURLs
+}
+
+// Run() gets the list of active devURLs from the cemanager for the
+// specified environment and outputs info to stdout.
+func (cmd urlsCmd) Run(fl *pflag.FlagSet) {
+	devURLs := urlList(fl)
+
```
Russtopia, line 184: Factored out the main fetch logic so it can be used by `del` and `create` subcommands as well.

cmd/coder/urls.go
```diff
@@ -14,13 +14,120 @@ import (
...
+	}
+}
+
+// Run() creates or updates a devURL, specified by env ID and port
+// (fl.Arg(0) and fl.Arg(1)), with access level (fl.Arg(2)) on
+// the cemanager.
+func (cmd createURLCmd) Run(fl *pflag.FlagSet) {
+	var envName = fl.Arg(0)
+	var port = fl.Arg(1)
+	var access = fl.Arg(2)
```
wbobeirne, line 55: We should provide a default value of private, and allow this to be optional.

We should also capitalize it for them, `public` or `org` should be valid input too.

Russtopia: Sounds good

Russtopia: I also wondered if I should do validation of the port arg here. Entering nonsensical value like 'fred' or '9999999' gives the server-side error, so do we care?

```$ ./coder create russ-env1 9999999 PRIVATE
Adding devurl for port 9999999
2020-06-25 08:39:37 ERROR	upsert devurl: http://localhost:8080/api/environments/5eecd3a0-c99dbfabfe3a9833e7172c5b/devurls
HTTP/1.1 400 Bad Request
Connection: close
Content-Length: 58
Coder-Version: 1.7.0-380-g7f5cbe16-ecf4366dc7
Content-Type: application/json
Date: Thu, 25 Jun 2020 15:39:20 GMT
Server: istio-envoy
Vary: Origin, Accept-Encoding
X-Envoy-Decorator-Operation: ingress-nginx.russtopia-rlmfuntoo.svc.cluster.local:80/*
X-Envoy-Upstream-Service-Time: 15
X-Powered-By: Express
```

Russtopia:
I still think the subcommand names `create` and `del` are too generic. If we enable more command-line operations via coder-cli such as managing organizations and users, these would be ambiguous.

How about `addurl` and `delurl` instead?

cmd/coder/urls.go
```diff
@@ -14,13 +15,124 @@ import (
...
+	if envName == "" {
+		exitUsage(fl)
+	}
+
+	if access == "" {
+		access = "PRIVATE"
+	}
+
+	access = strings.ToUpper(access)
+
```
Russtopia, line 67: Default PRIVATE if user gives no access level; and accept lowercase keywords

cmd/coder/sync.go
```diff
@@ -79,9 +75,7 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 		err = s.Run()
 	}
 
-	if err == NoRsync {
-		flog.Fatal("no compatible rsync present on remote machine")
-	} else {
+	if err != nil {
```
Russtopia, line 78: As seconded by @scsmithr we don't want to be checking errors this way; logic moved to internal/sync.go and now checks the process exit codes instead.

internal/sync/sync.go
```diff
@@ -66,6 +72,15 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
...
 	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == ERR_RSYNC_INCOMPAT {
+				flog.Error("no compatible rsync present on remote machine")
+			} else if exitError.ExitCode() == ERR_RSYNC_STREAM {
+				flog.Error("error in rsync protocol datastream (no installed remote rsync?)")
+			} else {
+				flog.Error("sync: %v", err)
+			}
+		}
```
Russtopia, line 83: use exec.ExitError to obtain and check actual OS exit codes

cmd/coder/urls.go
```diff
@@ -14,13 +15,124 @@ import (
...
+func (cmd createURLCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "create",
+		Usage: "<env name> <port> [<access>]",
+		Desc: `create/update a devurl for external access
+  <access> is one of [PRIVATE | ORG | AUTHED | PUBLIC]`,
+	}
+}
+
+// Run() creates or updates a devURL, specified by env ID and port
```
scsmithr, line 50: You can omit the `()`. Doc comments should start with just the identifier.

cmd/coder/urls.go
```diff
@@ -14,13 +15,124 @@ import (
...
+// (fl.Arg(0) and fl.Arg(1)), with access level (fl.Arg(2)) on
+// the cemanager.
+func (cmd createURLCmd) Run(fl *pflag.FlagSet) {
+	var envName = fl.Arg(0)
+	var port = fl.Arg(1)
+	var access = fl.Arg(2)
```
scsmithr, line 56: It's generally more idiomatic to put these in a var block.

```go
var (
    envName = fl.Arg(0)
    ...
)
```

cmd/coder/urls.go
```diff
@@ -14,13 +15,124 @@ import (
...
+
+	if envName == "" {
+		exitUsage(fl)
+	}
+
+	if access == "" {
+		access = "PRIVATE"
+	}
```
scsmithr, line 64: Should there also be a check for empty port?

cmd/coder/urls.go
```diff
@@ -29,7 +141,9 @@ func (cmd urlsCmd) Spec() cli.CommandSpec {
 	}
 }
 
-func (cmd urlsCmd) Run(fl *pflag.FlagSet) {
+// urlList() returns the list of active devURLs from the cemanager server.
+
+func urlList(fl *pflag.FlagSet) []DevURL {
```
scsmithr, line 146: Accepting a flag set makes it difficult to understand what this function actually requires. I'd suggest passing `envName` in instead.

internal/entclient/request.go
```diff
@@ -8,6 +8,9 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// request() makes an HTTP request of the specified method, returning the
+// *http.Response and any error.
+
```
scsmithr, line 13: There shouldn't be an empty line between the comment and the function.

internal/entclient/url.go
```diff
@@ -0,0 +1,45 @@
+package entclient
+
+import (
+	"fmt"
+	"net/http"
+)
+
+func (c Client) DelDevURL(envID, urlID string) error {
+	reqString := "/api/environments/%s/devurls/%s"
+	reqUrl := fmt.Sprintf(reqString, envID, urlID)
```
scsmithr, line 10: `reqURL` is more idiomatic.

internal/entclient/url.go
```diff
@@ -0,0 +1,45 @@
+package entclient
```
scsmithr, line 1: This file might be better named `devurl.go` to more accurately describe its purpose.

internal/sync/sync.go
```diff
@@ -41,6 +41,12 @@ type Sync struct {
 	Client *entclient.Client
 }
 
+// See https://lxadm.com/Rsync_exit_codes#List_of_standard_rsync_exit_codes.
+const (
+	ERR_RSYNC_INCOMPAT = 2
+	ERR_RSYNC_STREAM   = 12
```
scsmithr, line 47: Constants in go follow the same casing as other variables/functions. 

And I'd suggest naming these like `rsyncExitCodeIncompat`, since these aren't go errors.

internal/sync/sync.go
scsmithr, line 83: Instead of logging the error here, I would prefer that the error is wrapped.

So something like:
```go
if exitError, ok := err.(*exec.ExitError); ok {
    if exitError.ExitCode() == ERR_RSYNC_INCOMPAT {
        return xerrors.Errorf("no compatible rsync present on remote machine: %w", err)
    }
    ...
```

cmd/coder/main.go
```diff
@@ -37,6 +37,8 @@ func (r *rootCmd) Subcommands() []cli.Command {
 		&shellCmd{},
 		&syncCmd{},
 		&urlsCmd{},
+		&createURLCmd{},
+		&delURLCmd{},
```
scsmithr, line 41: These shouldn't be top level, they should be subcommands of `urlCmd`.

cmd/coder/urls.go
```diff
@@ -14,13 +15,124 @@ import (
...
+
+func isAccessLevelValid(level string) bool {
+	_, ok := urlAccessLevel[level]
+	return ok
+}
+
+func (cmd createURLCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "create",
+		Usage: "<env name> <port> [<access>]",
```
scsmithr, line 44: Given that `access` is optional, I think it may make sense to have it as a flag. I think being able to optionally provide `--access=public` would be nice.

(See `syncCmd` for an example of how to register flags.)

cmd/coder/urls.go
Russtopia: Ah, I always forget which way it is, thanks :)

cmd/coder/urls.go
Russtopia: I can add, but found the cli package is pretty robust, as well as the APIs we hit. If we don't want to see any of the backend API errors though I can try to validate port more thoroughly here

cmd/coder/urls.go
Russtopia: Ah, yes I might not need coupling to the entire FlagSet

internal/entclient/request.go
Russtopia: I did this on purpose for the two in this file, mostly as comments to myself or someone like me first seeing the code. I left the blank line so go doc wouldn't attach to the funcs (and go linters wouldn't complain about documenting a private func).

But I can just remove the comments if you like.

internal/entclient/url.go
Russtopia: Sure, will do

internal/sync/sync.go
Russtopia: Gah, old C habit again. Thanks. I'll follow the names of the error vars they replaced in the old code

internal/sync/sync.go
Russtopia: Ah, OK sure

cmd/coder/main.go
Russtopia: Oops, all right. Aha, that would mean users invoke as

`$ coder urls create <env_id> <port> <access>` ?

If so, I get now why more specific command tags weren't in the proposal.

cmd/coder/urls.go
Russtopia: Sure, if we want that style I'll switch it up.

cmd/coder/urls.go
Russtopia: Added port checks

cmd/coder/urls.go
Russtopia: I made it just --a, do you want it fully-spelled out ie., --access ?

wbobeirne: Single char arguments are usually single dashed, `-a` with a double dashed full word argument, `--access`

cmd/coder/urls.go
```diff
@@ -14,36 +16,163 @@ import (
...
+
+func (sub createSubCmd) Spec() cli.CommandSpec {
 	return cli.CommandSpec{
-		Name:  "urls",
-		Usage: "<env name>",
-		Desc:  "get all development urls for external access",
+		Name:  "create",
+		Usage: "<env name> <port> [--a access]",
+		Desc:  "create/update a devurl for external access",
 	}
```
Russtopia, line 69: I made it just --a, do you want it fully-spelled out ie., --access ?

cmd/coder/urls.go
```diff
@@ -14,36 +16,163 @@ import (
 )
 
 type urlsCmd struct{}
+type createSubCmd struct {
+	access string
+}
+type delSubCmd struct{}
```
scsmithr, line 22: Nitpicky, but these should be moved closer to their associated methods.

cmd/coder/urls.go
scsmithr: I'd prefer `--acess` for the flag. If you also want the `-a` shorthand, we could use [fl.StringVarP](https://godoc.org/github.com/spf13/pflag#StringVarP)

cmd/coder/urls.go
```diff
@@ -14,36 +16,163 @@ import (
...
+}
+
+func portIsValid(port string) bool {
+	p, err := strconv.ParseUint(port, 10, 16)
+	if p < 1 {
+		// port 0 means 'any free port', which we don't support
+		err = strconv.ErrRange
+	}
+	if err != nil {
+		fmt.Println("bad port")
```
scsmithr, line 47: I think we can do "Invalid port" here to make it consistent with `accessLevelIsValid`

PR #69, opened 2020-07-01
Fix SHELL environment variable

cmoog:


cmd/coder/shell.go
```diff
@@ -79,22 +80,18 @@ func (cmd *shellCmd) Run(fl *pflag.FlagSet) {
...
 		ctx     = context.Background()
 	)
 
-	var args []string
-	if command != "" {
-		args = fl.Args()[2:]
+	args := []string{"-c"}
+	if fl.Arg(1) == "" {
+		// Bring user into shell if no command is specified.
+		args = append(args, "export SHELL=$(getent passwd $(whoami) | awk -F: '{ print $7 }'); $SHELL")
```
Russtopia, line 89: Seems reasonable we can assume awk is present in any base Linux image but would we need to handle it not being there?

PR #71, opened 2020-07-01
Verify rsync protocol version match prior to proceeding - Rebased on current master.

### LINKED ISSUE #65: Check that rsync protocol versions match before initiating transfer
So we can give the user a better error message than Exit Status 2

stephenwithav:
`sync.Version()` proceeds with a warning that there may be a version
mismatch if it timesout.

`syncCmd.version` assumes rsync is present in the path.

`wsep` integration is currently unverified.  Verification is next.

Potentially closes #65.

### DIFF
--- a/cmd/coder/sync.go
+++ b/cmd/coder/sync.go
@@ -33,7 +33,7 @@ func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 	fl.BoolVarP(&cmd.init, "init", "i", false, "do initial transfer and exit")
 }
 
-// Returns local rsync protocol version as a string.
+// version returns local rsync protocol version as a string.
 func (_ *syncCmd) version() string {
 	cmd := exec.Command("rsync", "--version")
 	out, err := cmd.CombinedOutput()
--- a/cmd/coder/sync.go
+++ b/cmd/coder/sync.go
@@ -1,7 +1,12 @@
 package main
 
 import (
+	"bufio"
+	"errors"
+	"fmt"
+	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -27,6 +32,32 @@ func (cmd *syncCmd) Spec() cli.CommandSpec {
 
 func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 	fl.BoolVarP(&cmd.init, "init", "i", false, "do initial transfer and exit")
+}
+
+// See https://lxadm.com/Rsync_exit_codes#List_of_standard_rsync_exit_codes.
+var IncompatRsync = errors.New("rsync: exit status 2")
+var StreamErrRsync = errors.New("rsync: exit status 12")
+
+// Returns local rsync protocol version as a string.
+func (s *syncCmd) version() string {
+	cmd := exec.Command("rsync", "--version")
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if err := cmd.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	r := bufio.NewReader(stdout)
+	if err := cmd.Wait(); err != nil {
+		log.Fatal(err)
+	}
+
+	versionString := strings.Split(r.ReadLine(), "protocol version ")
+
+	return versionString[1]
 }
 
 func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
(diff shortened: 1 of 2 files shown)


stephenwithav:
This builds locally fine.

The CI bug is inaccessible to me.  Any tips, @scsmithr?

cmd/coder/sync.go
```diff
@@ -29,6 +34,37 @@ func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
...
+	}
+
+	firstLine, err := r.ReadString('\n')
+	if err != nil {
+		return ""
+	}
+
+	versionString := strings.Split(firstLine, "protocol version ")
+
+	return versionString[1]
```
scsmithr, line 65: The way this is laid out, `version` will always return an empty string. Reading from `r` will always error, because the pipe will be closed before we start reading from the command.

I would suggest replacing reading the output of the command with:
```
out, err := cmd.CombinedOutput()
if err != nil {
    log.Fatal(err)
}
```

And I would replace reading the first line with:
```
firstLine, err := bytes.NewBuffer(out).ReadString('\n')
if err != nil {
    log.Fatal(err)
}
```

internal/sync/sync.go
```diff
@@ -261,6 +263,46 @@ const (
...
+
+	execer := wsep.RemoteExecer(conn)
+	process, err := execer.Start(ctx, wsep.Command{
+		Command: "rsync",
+		Args:    []string{"--version"},
+	})
+	if err != nil {
+		return "", err
+	}
+	r := bufio.NewReader(process.Stdout())
```
scsmithr, line 286: We have the same issue here with the reader never reading anything.

I think the best solution here would be to replace this reader with:
```
buf := &bytes.Buffer{}
io.Copy(buf, process.Stdout())
```

And `buf` can be used down below to read the first line.

cmd/coder/sync.go
stephenwithav: Will do.

internal/sync/sync.go
stephenwithav: Will do.

stephenwithav: Do you want the remote check to `log.Fatal` fail if we can't retrieve it?

scsmithr: I think I prefer the behavior you had, where we just return the error and warn the user.

stephenwithav: The remote check still has that.

internal/sync/sync.go
```diff
@@ -261,6 +263,47 @@ const (
...
+	execer := wsep.RemoteExecer(conn)
+	process, err := execer.Start(ctx, wsep.Command{
+		Command: "rsync",
+		Args:    []string{"--version"},
+	})
+	if err != nil {
+		return "", err
+	}
+	buf := &bytes.Buffer{}
+	go io.Copy(buf, process.Stdout())
```
scsmithr, line 287: I would remove the `go` here. `io.Copy` will will terminate as expected once the command completes.

Having a goroutine here can be racy, because there's no guarantees that it runs or completes the copy before we start reading from the buffer.

stephenwithav: Done.

internal/sync/sync.go
```diff
@@ -261,6 +263,47 @@ const (
...
+	})
+	if err != nil {
+		return "", err
+	}
+	buf := &bytes.Buffer{}
+	io.Copy(buf, process.Stdout())
+
+	err = process.Wait()
+	if code, ok := err.(wsep.ExitError); ok {
+		return "", fmt.Errorf("Version check exit status: %v", code)
```
scsmithr, line 291:
```suggestion
		return "", fmt.Errorf("version check exit status: %v", code)
```

internal/sync/sync.go
```diff
@@ -261,6 +263,47 @@ const (
...
+	}
+	if err != nil {
+		return "", fmt.Errorf("Server version mismatch")
```
scsmithr, line 294: I'd prefer returning the error here instead.

internal/sync/sync.go
```diff
@@ -261,6 +263,47 @@ const (
...
+	if err != nil {
+		return "", err
+	}
+	buf := &bytes.Buffer{}
+	io.Copy(buf, process.Stdout())
+
+	err = process.Wait()
+	if _, ok := err.(wsep.ExitError); ok {
+		return "", err
+	}
```
scsmithr, line 292: We may as well remove this branch.

cmd/coder/sync.go
```diff
@@ -29,6 +33,23 @@ func (cmd *syncCmd) RegisterFlags(fl *pflag.FlagSet) {
 	fl.BoolVarP(&cmd.init, "init", "i", false, "do initial transfer and exit")
 }
 
+// Returns local rsync protocol version as a string.
```
scsmithr, line 36: Apologies for missing this earlier in the review. These comments should start with the function/method name. So this would be "version returns local rsync protocol version as a string."

(And same for the other `Version` method)

stephenwithav: No problem.  I appreciate you teaching me the style you prefer prior to the interview next week so if I get the job, I'm already style trained.

PR #72, opened 2020-07-02
Replace rsync sync with mutagen

ammario:
- [ ] Clean up UX around monitoring syncs
- [ ] Add automatic agent installation
- [ ] General code cleanup


cmoog:
My understanding is that there are no plans to move forward with merging this.

PR #74, opened 2020-07-15
Add custom prefix devurl

Russtopia:
This implements  https://github.com/cdr/enterprise/issues/4411
***NOTE: this is intended for post 1.9 release. It involves a breaking API change to the devurl API and should be released in coordination with https://github.com/cdr/enterprise/pull/4381***

### What This Does

* Adds new `urls create` subcommand option `--name <label | "">`
* Adds new `urls ls` command and `-o [human | json]` options following syntax of the `users ls` command
* Some error message output is modified to be more consistent (fmt.Prinf() vs. coder.com/flog)
* ***Propose new `edit` urls subcommand which is just an alias for `create`; just for usage style***
* ~~Propose to make `create` single-purpose (only create a new devurl) such that `edit` is the only way to modify an existing devurl~~
* Scaffolding for integration tests of devurl operations

Changes for compatibility with https://github.com/cdr/enterprise/pull/4381
* REST APIs to manipulate devurls have the `port` argument changed to int type

### Testing
* Use https://github.com/cdr/enterprise/pull/4381 deployment to test against to ensure API compatibility.

### Usage Examples
* TODO


cmd/coder/urls.go
```diff
@@ -35,47 +36,52 @@ var urlAccessLevel = map[string]string{
...
-	if p < 1 {
+	if p < 1 || p > 65535 {
 		// port 0 means 'any free port', which we don't support
 		err = strconv.ErrRange
 	}
 	if err != nil {
-		fmt.Println("Invalid port")
+		flog.Error("Invalid port")
 	}
 	return err == nil
```
Russtopia, line 46: In cdr/enterprise the convention appears to prefer multiple return points where errors are first detected; if we want to follow that for the coder-cli repo I can update everything I touched in this PR to use that style.

cmd/coder/urls.go
```diff
@@ -35,47 +36,52 @@ var urlAccessLevel = map[string]string{
...
 }
 
 func accessLevelIsValid(level string) bool {
 	_, ok := urlAccessLevel[level]
 	if !ok {
-		fmt.Println("Invalid access level")
+		flog.Error("Invalid access level")
```
Russtopia, line 52: Changed Printf() and Println() occurrences to use flog, as that appeared to be the desired way in @ammario previous work

cmd/coder/urls.go
```diff
@@ -35,47 +36,52 @@ var urlAccessLevel = map[string]string{
...
 }
 
 func (sub *createSubCmd) RegisterFlags(fl *pflag.FlagSet) {
 	fl.StringVarP(&sub.access, "access", "a", "private", "[private | org | authed | public] set devurl access")
+	fl.StringVarP(&sub.urlname, "name", "n", "", "devurl name")
 }
 
 func (sub createSubCmd) Spec() cli.CommandSpec {
 	return cli.CommandSpec{
 		Name:  "create",
```
Russtopia, line 69: I wonder if we should define a synonym `update` for user-friendliness. `create` works for both insert and update of devurls.

cmd/coder/urls.go
```diff
@@ -84,26 +90,35 @@ func (sub createSubCmd) Run(fl *pflag.FlagSet) {
 	if !portIsValid(port) {
 		exitUsage(fl)
 	}
+	portNum, _ := strconv.Atoi(port)
```
Russtopia, line 93: Ignoring err return here as port was just validated above. Can make portIsValid() return validated portNum instead if you'd prefer (probably would rename to validatePort() then)

```
portNum, err := validatePort(port) {
        if err != nil { exitUsage(f1) }
}
```

cmd/coder/urls.go
```diff
@@ -35,47 +36,52 @@ var urlAccessLevel = map[string]string{
...
 func (sub createSubCmd) Spec() cli.CommandSpec {
 	return cli.CommandSpec{
 		Name:  "create",
-		Usage: "<env name> <port> [--access <level>]",
+		Usage: "<env name> <port> [--access <level>] [--name <name>]",
 		Desc:  "create/update a devurl for external access",
 	}
 }
 
+var devURLNameValidRx = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9]{0,63}$")
```
Russtopia, line 75: Will add doc comment to this

internal/entclient/devurl.go
```diff
@@ -24,22 +32,56 @@ func (c Client) DelDevURL(envID, urlID string) error {
...
+	}
 	if err != nil {
 		return err
 	}
+	return nil
+}
 
-	if res.StatusCode != http.StatusOK {
-		return bodyError(res)
-	}
```
Russtopia, line 42: I thought it might be nicer to give a plain English error message rather than just dump the entire HTTP error response to console.

cmd/coder/urls.go
```diff
@@ -84,26 +90,35 @@ func (sub createSubCmd) Run(fl *pflag.FlagSet) {
...
 
 	access = strings.ToUpper(sub.access)
 	if !accessLevelIsValid(access) {
 		exitUsage(fl)
 	}
 
+	name = sub.urlname
+	if name != "" && !devURLNameValidRx.MatchString(name) {
+		flog.Error("update devurl: name must be < 64 chars in length, begin with a letter and only contain letters or digits.")
+		return
```
Russtopia, line 103: Given our target audience is developers, would it better (since shorter) to just say

`"update devurl: name must conform to regex ^[a-zA-Z][a-zA-Z0-9]{0,63}$"`
or
`"update devurl: name must be <= 64 chars and conform to regex ^[a-zA-Z][a-zA-Z0-9]"`
?

cmd/coder/urls.go
```diff
@@ -35,47 +36,66 @@ var urlAccessLevel = map[string]string{
 
 func portIsValid(port string) bool {
 	p, err := strconv.ParseUint(port, 10, 16)
-	if p < 1 {
+	if p < 1 || p > 65535 {
```
Russtopia, line 39: I should probably just make this `if p < 1 || p > math.MaxUint16`

cmd/coder/urls.go
```diff
@@ -35,47 +36,66 @@ var urlAccessLevel = map[string]string{
...
+type editSubCmd struct {
+	createSubCmd
+}
+
+func (sub *editSubCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "edit",
+		Usage: "<env name> <port> [--access <level>] [--name <name>]",
+		Desc:  "create/update a devurl for external access",
+	}
```
Russtopia, line 70: This adds a synonym `edit` for `create` purely for usage style, this way users can invoke in ways that show their intention eg:

`coder urls create ...` then
`coder urls edit ...`

cmd/coder/urls.go
```diff
@@ -193,7 +223,7 @@ func urlList(envName string) []DevURL {
 	}
 
 	if len(devURLs) == 0 {
-		fmt.Printf("no dev urls were found for environment: %s\n", envName)
+		flog.Error("no dev urls were found for environment: %s", envName)
```
Russtopia, line 226: Unrelated to this review: It's a bit confusing to the user we *always* log an error in red to the console if there are no devurls here. This function is used to get the list of existing devurls, which can legitimately be empty. I'll look at moving the error handling up to where we know whether to expect an empty list or not, if that makes sense.

cmd/coder/urls.go
```diff
@@ -35,47 +37,66 @@ var urlAccessLevel = map[string]string{
 
 func portIsValid(port string) bool {
 	p, err := strconv.ParseUint(port, 10, 16)
-	if p < 1 {
+	if p < 1 || p > math.MaxUint16 {
```
coadler, line 40: It's not possible for this function to return a value over the provided bitsize anyways, so this check isn't doing much. See https://play.golang.org/p/8aTqC70r7Ou

I think this error handling here is a little backwards. We should be checking `err` before checking `p` since `p` will be 0 if `err != nil`. If we're checking that `p < 1` before `err` then we'll be returning the wrong error.

internal/entclient/devurl.go
coadler: Usually the API returns helpful error text and resolution advice. If the responses aren't helpful, maybe we should make them that way?

Russtopia: Definitely. I'll look into it. Would be nice if the coder-cli didn't have to do its own explanations and could re-use what we (hopefully!) give back to the web dashboard etc.

cmd/coder/urls.go
Russtopia: OK. Duh, yes I see now I'd already called ParseUint() so it has to be in range :)

Russtopia:
In general: how do we feel about the `create` and `edit` subcommands (versus just `create` in master)?
I feel it's more sensible from a user standpoint.. and non-intuitive that `create` could also *update* an existing devurl. If no one has objections, I'd like to make `create` actually refuse to update an existing devurl, and let `edit` do that.

cmd/coder/urls.go
```diff
@@ -117,15 +147,16 @@ func (sub delSubCmd) Spec() cli.CommandSpec {
 	}
 }
 
-// devURLID returns the ID of a devURL, given the env name and port.
-// ("", false) is returned if no match is found.
-func devURLID(port string, urls []DevURL) (string, bool) {
+// devURLID returns the ID of a devURL, given the env name and port
+// from a list of DevURL records.
+// (false, "") is returned if no match is found.
+func devURLID(port int, urls []DevURL) (bool, string) {
```
scsmithr, line 153: It's more idiomatic to return (string, bool).

cmd/coder/sync.go
```diff
@@ -99,7 +99,7 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 	if rsyncErr != nil {
 		flog.Info("Unable to determine remote rsync version.  Proceeding cautiously.")
 	} else if localVersion != remoteVersion {
-		flog.Fatal(fmt.Sprintf("rsync protocol mismatch. %s.", localVersion, rsyncErr))
+		flog.Fatal(fmt.Sprintf("rsync protocol mismatch. %s (%v).", localVersion, rsyncErr))
```
Russtopia, line 102: Seems this was an error that slipped into master. fmt.Sprintf() arg mismatch.

cmd/coder/urls.go
Russtopia: OK will swap.

internal/entclient/devurl.go
Russtopia: If it's OK, I'll leave deeper changes to error message to another PR since it'll involve possible changes to cdr/enterprise as well.

Russtopia:
> Regarding the create/edit subcommands, I'd prefer to either keep how you have it right now in the pr, or go back to having just `create`. Personally, I like that it's idempotent(ish) and I'd find it annoying to have to call `edit` in the case of me forgetting that a devurl already exists. I like my CLIs to just do what I mean.

Sure. Since @fuskovic  added an alias ability to the cli recently, I left in "edit", but one can just use `create` for both ops.

cmd/coder/urls.go
```diff
@@ -33,55 +55,110 @@ var urlAccessLevel = map[string]string{
...
+
+	switch sub.outputFmt {
+	case "human":
+		w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', tabwriter.TabIndent)
+		for _, devURL := range devURLs {
+			fmt.Fprintf(w, "%s\t%d\t%s\n", devURL.URL, devURL.Port, devURL.Access)
+		}
+		err := w.Flush()
+		if err != nil {
+			flog.Fatal("failed to flush writer: %v", err)
```
cmoog, line 99: Checkout `internal/x/xtabwriter`. Should help make this section much cleaner. `secrets.go` has a good example.

cmoog: I've been thinking of adding support for a struct tag `human:"omit"` such that you could only print the fields that you want...

ci/integration/devurls_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// // == devurl del should function
+	// c.Run(ctx, `coder urls del env1 3002`).Assert(t,
+	// 	tcli.Success())
+
+	// // == removed devurl should no longer be there
+	// c.Run(ctx, `coder urls ls env1 -o json | jq -c '.[] | select( .name == "foobar")'`).Assert(t,
+	// 	tcli.Error(),
+	// 	jsonUnmarshals(&durl))
+
+}
```
Russtopia, line 94: Speculative integration tests -- we need envs to be created in the test CI environment, or to add `env create` ability to coder-cli itself, before we can do any of these.

internal/entclient/devurl.go
```diff
@@ -5,18 +5,33 @@ import (
...
 )
 
+// DevURL is the parsed json response record for a devURL from cemanager
+type DevURL struct {
+	ID     string `json:"id"`
+	URL    string `json:"url"`
+	Port   int    `json:"port"`
+	Access string `json:"access"`
+	Name   string `json:"name"`
+}
```
Russtopia, line 15: Moved in here, where I presume it should be (like `entclient.User`)

cmd/coder/urls.go
```diff
@@ -33,55 +55,110 @@ var urlAccessLevel = map[string]string{
...
 }
 
+type listSubCmd struct {
+	outputFmt string
+}
+
+// Run gets the list of active devURLs from the cemanager for the
+// specified environment and outputs info to stdout.
+func (sub listSubCmd) Run(fl *pflag.FlagSet) {
+	envName := fl.Arg(0)
```
cmoog, line 88: Validate that this is nonempty before continuing.

cmd/coder/urls.go
cmoog, line 99: Use `requireSuccess(err, "failed to flush writer: %v", err)` for brevity. The same could be said a few other places.

cmd/coder/urls.go
```diff
@@ -33,55 +55,110 @@ var urlAccessLevel = map[string]string{
...
+
+func (sub *listSubCmd) RegisterFlags(fl *pflag.FlagSet) {
+	fl.StringVarP(&sub.outputFmt, "output", "o", "human", "output format (human | json)")
+}
+
+func (sub *listSubCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "ls",
+		Usage: "<env> <flags>",
+		Desc:  "list all devurls",
```
cmoog, line 119: `list all devurls for a given environment`

cmd/coder/urls.go
```diff
@@ -90,20 +167,28 @@ func (sub createSubCmd) Run(fl *pflag.FlagSet) {
 		exitUsage(fl)
 	}
 
+	name = sub.urlname
+	if name != "" && !devURLNameValidRx.MatchString(name) {
+		flog.Error("update devurl: name must be < 64 chars in length, begin with a letter and only contain letters or digits.")
+		return
```
cmoog, line 173: This should be fatal.

cmd/coder/urls.go
```diff
@@ -90,20 +167,28 @@ func (sub createSubCmd) Run(fl *pflag.FlagSet) {
...
-		fmt.Printf("Adding devurl for port %v\n", port)
-	}
-
-	err := entClient.UpsertDevURL(env.ID, port, access)
-	if err != nil {
-		flog.Error("upsert devurl: %s", err.Error())
+		flog.Info("Adding devurl for port %v", port)
+		err := entClient.InsertDevURL(env.ID, portNum, name, access)
+		if err != nil {
+			flog.Error("insert devurl: %s", err.Error())
```
cmoog, line 190: Should be fatal here as well.

cmd/coder/urls.go
```diff
@@ -90,20 +167,28 @@ func (sub createSubCmd) Run(fl *pflag.FlagSet) {
...
 	env := findEnv(entClient, envName)
 
-	_, found := devURLID(port, urlList(envName))
+	urlID, found := devURLID(portNum, urlList(envName))
 	if found {
-		fmt.Printf("Updating devurl for port %v\n", port)
+		flog.Info("Updating devurl for port %v", port)
+		err := entClient.UpdateDevURL(env.ID, urlID, portNum, name, access)
+		if err != nil {
+			flog.Error("update devurl: %s", err.Error())
```
cmoog, line 184: Should be fatal.

cmd/coder/urls.go
```diff
@@ -137,35 +223,27 @@ func (sub delSubCmd) Run(fl *pflag.FlagSet) {
...
-		fmt.Printf("Deleting devurl for port %v\n", port)
+		flog.Info("Deleting devurl for port %v", port)
 	} else {
 		flog.Fatal("No devurl found for port %v", port)
 	}
 
-	err := entClient.DelDevURL(env.ID, urlID)
+	err = entClient.DelDevURL(env.ID, urlID)
 	if err != nil {
 		flog.Error("delete devurl: %s", err.Error())
```
cmoog, line 243: should be fatal.

cmd/coder/urls.go
```diff
@@ -192,29 +270,5 @@ func urlList(envName string) []DevURL {
...
-	for _, devURL := range devURLs {
-		fmt.Fprintf(w, "%s\t%s\t%s\n", devURL.URL, devURL.Port, devURL.Access)
-	}
-	w.Flush()
-}
-
-func (cmd *urlsCmd) Subcommands() []cli.Command {
-	return []cli.Command{
-		&createSubCmd{},
-		&delSubCmd{},
```
cmoog, line 218: can we make this command `rm` for consistency?

cmd/coder/urls.go
Russtopia: Could be cool. What if users wanted more verbose vs. brief output -- would something like `human:"verbose"` and `human:"always"` or something allow multiple verbosity levels?

cmd/coder/urls.go
Russtopia: OK. Currently it'll output `[]` for json and nothing for tabbed. You want no output for both or a msg such as `No devurls defined` for human output?

cmd/coder/urls.go
Russtopia: As discussed via Slack I'll leave this for now but once xtabwriter supports marking fields to suppress output I will convert it to use that. (Just didn't want to output the `name` field since it's redundant, being part of the devurl itself).

Russtopia: Ah nice. Hadn't noticed we had that in here.

cmd/coder/urls.go
```diff
@@ -137,35 +223,27 @@ func (sub delSubCmd) Run(fl *pflag.FlagSet) {
 		exitUsage(fl)
 	}
 
-	if !portIsValid(port) {
+	portNum, err := validatePort(port)
+	if err != nil {
 		exitUsage(fl)
 	}
```
Russtopia, line 229: Hmm. I wonder if this and `accessLevelIsValid()` should then just call `exitUsage()` and `log.Fatal()` themselves rather than `log.Error()` and relying on the caller to exit. What's your preference on exit/fatal from 'helper' funcs?

cmd/coder/urls.go
Russtopia: Changed for now to output nothing for empty devURLs

cmd/coder/urls.go
Russtopia: Done.

cmd/coder/urls.go
```diff
@@ -137,33 +219,23 @@ func (sub delSubCmd) Run(fl *pflag.FlagSet) {
...
-}
-
-func (cmd urlsCmd) Spec() cli.CommandSpec {
-	return cli.CommandSpec{
-		Name:  "urls",
-		Usage: "<env name>",
-		Desc:  "get all development urls for external access",
-	}
+	err = entClient.DelDevURL(env.ID, urlID)
+	requireSuccess(err, "delete devurl: %s", err)
```
Russtopia, line 238: Am I just being thick here, passing err twice? I actually want the HTTP error response to be output so it seems that I need to.

cmoog: Yes that's how I intended it to be used-- not the best looking I'll admit.

cmoog:
@Russtopia Anything blocking this from merging?

Issue #76, opened 2020-07-27
Create foundation for performing bulk actions


cmoog:
Allow admins to give a `--user` flag that specifies which users's resources should be the target of crud operations. Example:
```
$ coder secrets --user charlie@coder.com create mysql-password --from-literal 123password
```

Related: https://github.com/cdr/enterprise/pull/4497

Issue #63, opened 2020-06-29
Integration tests for coder-cli


Russtopia:
As coder-cli runs against the backend (cdr/enterprise) and depends on its APIs, it would make sense to do some tests of the coder-cli tool itself as part of the main integration checks.

Major features could be tested to ensure they still function, eg:

* sync
* sh
* devurls
  - add
  - update
  - list
  - delete

wbobeirne:
I had also considered this in https://github.com/cdr/enterprise/issues/4230 (Probably should have opened the issue here.) Some upsides of having tests in enterprise is that:
* We'll test changes as the API changes, so we can catch API breaks as a part of the PR in enterprise, instead of tests randomly break in here
* We'll be able to re-use the data scaffolding code for frontend e2e tests, so that you can independently test each method instead of having to test CRUD as one whole flow

Russtopia:
I can close this one and move comments to #4230 if you'd prefer -- let me know, thanks

cmoog:
Closing as the `coder-cli` work is done. The remaining phases involve configuring enterprise namespace deployments and pointing the integrations tests towards mocked APIs.

PR #77, opened 2020-07-27
Add list users command

cmoog:
Related to #76 

## Example Usage
```
$ coder users ls -o json | jq -c '.[] | select( .username == "charlie")' 
{"id":"5e8decf9-563f9fb3a7c02f6e0eafa422","email":"charlie@coder.com","username":"charlie","name":"Charlie","created_at":"2020-04-08T15:25:45.419609Z"}

$ coder users ls -o human | grep charlie
5e8decf9-563f9fb3a7c02f6e0eafa422        charlie@coder.com           charlie                                         Charlie   2020-04-08 15:25:45.419609 +0000 UTC
```

### DIFF
--- a/cmd/coder/users.go
+++ b/cmd/coder/users.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/spf13/pflag"
+
+	"go.coder.com/cli"
+	"go.coder.com/flog"
+)
+
+type usersCmd struct {
+}
+
+func (cmd usersCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "users",
+		Usage: "[subcommand] <flags>",
+		Desc:  "interact with user accounts",
+	}
+}
+
+func (cmd usersCmd) Run(fl *pflag.FlagSet) {
+	exitUsage(fl)
+}
+
+func (cmd *usersCmd) Subcommands() []cli.Command {
+	return []cli.Command{
+		&listCmd{},
+	}
+}
+
+type listCmd struct {
+	outputFmt string
+}
+
+func tabDelimited(data interface{}) string {
+	v := reflect.ValueOf(data)
+	s := &strings.Builder{}
+	for i := 0; i < v.NumField(); i++ {
+		s.WriteString(fmt.Sprintf("%s\t", v.Field(i).Interface()))
+	}
+	return s.String()
+}
+
+func (cmd *listCmd) Run(fl *pflag.FlagSet) {
+	entClient := requireAuth()
+
+	users, err := entClient.Users()
+	if err != nil {
+		flog.Fatal("failed to get users: %v", err)
+	}
+
+	switch cmd.outputFmt {
+	case "human":
+		w := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+		for _, u := range users {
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -17,3 +17,24 @@ Download the latest [release](https://github.com/cdr/coder-cli/releases):
 cd ~/go/bin
 tar -xvf ~/Downloads/coder-cli-linux-amd64.tar.gz
 ```
+
+## Usage
+
+```
+$ coder --help
+Usage: coder [subcommand] [flags]
```
exvuma, line 25: We shouldn't really maintain the output of commands in the README since it changes frequently. We should just list there is a --help flag and remove the output.

cmd/coder/users.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (cmd *listCmd) RegisterFlags(fl *pflag.FlagSet) {
+	fl.StringVarP(&cmd.outputFmt, "output", "o", "human", "specify the output format")
+}
+
+func (cmd *listCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "ls",
+		Usage: "<flags>",
+		Desc:  "list all users",
```
exvuma, line 91:
```suggestion
		Desc:  "list all users that the signed-in account has admin access to sitewide",
```

cmd/coder/users.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type usersCmd struct {
+}
+
+func (cmd usersCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "users",
+		Usage: "[subcommand] <flags>",
+		Desc:  "used to interact with user accounts",
```
exvuma, line 24:
```suggestion
		Desc:  "interact with user accounts",
```

cmd/coder/sync.go
```diff
@@ -99,7 +98,7 @@ func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
 	if rsyncErr != nil {
 		flog.Info("Unable to determine remote rsync version.  Proceeding cautiously.")
 	} else if localVersion != remoteVersion {
-		flog.Fatal(fmt.Sprintf("rsync protocol mismatch. %s.", localVersion, rsyncErr))
+		flog.Fatal("rsync protocol mismatch: local = %v, remote = %v", localVersion, rsyncErr)
```
exvuma, line 101: Why is this in this PR? just curious seems like it's own thing

README.md
cmoog: I find it very helpful from a user perspective in understanding the basic functionality within the first 10 seconds of interacting with the repo. I'll add a CI check as part of #63 .

cmd/coder/sync.go
cmoog: Was causing some linting errors so I figured I'd clean it up.

cmoog: Won't happen again once we get proper CI.

README.md
cmoog: On second thought, let's table this for now and revisit when the state of the documentation can be fully considered.

cmd/coder/users.go
cmoog: All users have perms to list other users, not just admin.

exvuma: It'd be nice to make this clear either way and use the help to tell us more than what the command already does e.g `users ls` says `lists users` but it's not clear what users

cmoog: list _all_ users

Issue #67, opened 2020-07-01
Support and create a version compatible with Windows


nathanpotter:

Russtopia:
If there's not a hard requirement to stick to vanilla Windows, porting would be eased by leveraging MSYS
https://packages.msys2.org/package/rsync

cmoog:
@nathanpotter Per your comment today about discouraging local ide workflows-- I'm thinking we should drop plans to support mutagen for now. It is only useful for local IDE workflows and requires SSH– unlike `coder sync`. 

Let's just hide `coder sync` behind a build flag for windows and replace it with an error message.

cmoog:
It looks like we're also blocked by our PTY library https://github.com/creack/pty

nathanpotter:
I agree^ one option is to just not support `coder sync` on windows and only build it into the tool for linux / darwin through go build flags / files

ammario:
@nathanpotter's solution is the best initial way forward

cmoog:
@nathanpotter @ammario totally agree. See https://github.com/cdr/coder-cli/issues/67#issuecomment-664026264. We're also blocked by `coder sh` seudo terminals.

nathanpotter:
Are we able to utilize a different pty library that supports the various platforms?

cmoog:
> Are we able to utilize a different pty library that supports the various platforms?

Unclear... still trying to find one.

PR #78, opened 2020-07-27
Hide sync behind build flag for windows

cmoog:
Is not strictly required for Windows users, but improves UX. 
Related: #67

### DIFF
--- a/internal/cmd/sync_unix.go
+++ b/internal/cmd/sync_unix.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package cmd
 
 import (
--- a/ci/steps/build.sh
+++ b/ci/steps/build.sh
@@ -29,8 +29,8 @@ build(){
 # Darwin builds do not work from Linux, so only try to build them from Darwin.
 # See: https://github.com/cdr/coder-cli/issues/20
 if [[ "$(uname)" == "Darwin" ]]; then
-	GOOS=linux build
 	CGO_ENABLED=1 GOOS=darwin build
+	GOOS=linux build
 	GOOS=windows GOARCH=386 build
 	exit 0
 fi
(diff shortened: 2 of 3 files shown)


cmd/coder/sync_windows.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type syncCmd struct {
+}
+
+func (cmd *syncCmd) Spec() cli.CommandSpec {
+	return cli.CommandSpec{
+		Name:  "sync",
+		Usage: "[local directory] [<env name>:<remote directory>]",
+		Desc:  "establish a one way directory sync to a remote environment",
```
deansheather, line 19: I'd put that it's unavailable for windows here too.

cmd/coder/sync_windows.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+}
+
+func (cmd *syncCmd) Run(fl *pflag.FlagSet) {
+	flog.Fatal(`"coder sync" is not available for Windows`)
```
deansheather, line 24: How hard is it for us to just support this? We just use rsync on both sides of the transfer AFAIK, so as long as the user installs rsync manually and has OpenSSH Client feature enabled on Windows it shouldn't require that much effort from us to support.

deansheather: Russ already mentioned this actually: https://packages.msys2.org/package/rsync

I think if we provided some quick setup instructions for getting rsync installed and working on Windows it'd be a great bonus.

cmoog: I don't have a strong opinion either way.  @nathanpotter thoughts?

internal/cmd/sync_unix.go
```diff
@@ -1,3 +1,5 @@
+// +build !windows
```
creack, line 1: The build tags are directives for the compiler, not comments. To indicate this, there should not be a space after `//`.

cmoog:
Outdated.

PR #79, opened 2020-07-27
Adds support for windows

### LINKED ISSUE #67: Support and create a version compatible with Windows

cmoog:
Closes #67


cmd/coder/resize_unix.go
```diff
@@ -0,0 +1,39 @@
+// +build !windows
+
+package main
```
deansheather, line 3: The resize files could potentially be put into the xterminal package since it makes more sense there.

PR #80, opened 2020-07-28
Initial setup for integration tests

cmoog:
Progress towards #63 https://github.com/cdr/enterprise/issues/4230

This PR introduces a new CLI testing framework modeled after [https://github.com/cdr/hat](https://github.com/cdr/hat).

The tentative name is `tcli`. Commands are executed inside docker containers to limit pollution of the host filesystem and to promote reproducibility. `tcli` uses a flexible and extensible `Assertion` API for asserting properties about the `CommandResult`.

View a sample usage of `tcli` [here](https://github.com/cdr/coder-cli/pull/80/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod&file-filters%5B%5D=.yaml#diff-6fdd64f1328ce60add8dd1eacfa4b463)

Given that `coder login` currently requires a browser, this PR uses a workaround approach by which a `session_token` is acquired through a manual call to the `/auth/basic/login` endpoint. The test then injects the credentials into their respective configuration files within the given container manually.


cmoog:
This is ready to merge.

ci/tcli/tcli.go
```diff
@@ -0,0 +1,328 @@
...
+	cmd.Args = append([]string{"docker", "exec", "-i", r.name}, cmd.Args...)
+
+	return &Assertable{
+		cmd:   cmd,
+		tname: command,
+	}
+}
+
+// HostRunner executes command tests on the host, outside of a container
+type HostRunner struct{}
```
scsmithr, line 133: Do you foresee us using `HostRunner` in the coder-cli tests?

ci/tcli/tcli.go
```diff
@@ -0,0 +1,328 @@
...
+
+	start := time.Now()
+	err := a.cmd.Run()
+	result.Duration = time.Since(start)
+
+	if exitErr, ok := err.(*exec.ExitError); ok {
+		result.ExitCode = exitErr.ExitCode()
+	} else if err != nil {
+		// TODO: handle this case better
+		result.ExitCode = -1
```
scsmithr, line 185: Could we fail the test here and log the error?

ci/tcli/tcli.go
cmoog: Not at the moment, but I do think it has a place in `tcli` as an escape hatch.

ci/tcli/tcli.go
cmoog: Yeah, that seems like the best option for now.

PR #81, opened 2020-07-30
Add secrets ls, add, view, rm commands

cmoog:
```
Usage: coder secrets [subcommand]

interact with secrets

Commands:
        ls      list all secrets
        view    view a secret
        create  insert a new secret
        rm      remove a secret
```

```
bash-3.2$ coder secrets create --name mysql-password --value 123

bash-3.2$ coder secrets ls
ID                                  Name             Value    Description   CreatedAt                              UpdatedAt                              
5f22365f-d7fd0f424e93930a9f8b9df1   mysql-password   ******                 2020-07-30 02:54:23.424945 +0000 UTC   2020-07-30 02:54:23.424945 +0000 UTC   

bash-3.2$ coder secrets view mysql-password
123

bash-3.2$ coder secrets rm mysql-password
2020-07-29 21:54:43 INFO        Successfully deleted secret "mysql-password"

bash-3.2$ coder secrets ls
2020-07-29 21:54:47 INFO        No secrets found
```

cc @wbobeirne @exvuma


internal/x/xtabwriter/tabwriter.go
```diff
@@ -0,0 +1,34 @@
+package xtabwriter
```
scsmithr, line 1: This is a great addition.

cmoog: Just wait till we get a struct tag in there
```go
type User struct {
  ID string `human:"hide"`
}
```

PR #89, opened 2020-08-03
Environment subcommands

cmoog:


cmoog:
Merged these commits into #86

PR #92, opened 2020-08-04
Add --user flag to "secrets" and "envs"

### LINKED ISSUE #76: Create foundation for performing bulk actions

cmoog:
## Samples
Setting a secret value for all users. Of course, were the admin to want a different value for each user, they would need to have some additional way of correlating the email to a secret prior to creating the Coder secret.
```bash
#!/bin/bash

# get all user emails
emails=$(coder users ls --output json | jq -c -r '.[] | .email')

for user in $emails
do
	# create a secret for that user
	coder secrets --user $user create mysql-password --from-literal password123 
done
```

Closes #76

### DIFF
--- a/cmd/coder/auth.go
+++ b/cmd/coder/auth.go
@@ -35,8 +35,10 @@ func newClient() (*entclient.Client, error) {
 		return nil, xerrors.Errorf("url misformatted: %v (try runing coder login)", err)
 	}
 
-	return &entclient.Client{
+	client := &entclient.Client{
 		BaseURL: u,
 		Token:   sessionToken,
-	}, nil
+	}
+
+	return client, nil
 }
--- a/cmd/coder/ceapi.go
+++ b/cmd/coder/ceapi.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"context"
+
 	"golang.org/x/xerrors"
 
 	"go.coder.com/flog"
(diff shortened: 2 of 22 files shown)


cmd/coder/auth.go
```diff
@@ -11,15 +11,15 @@ import (
 )
 
 // requireAuth exits the process with a nonzero exit code if the user is not authenticated to make requests
-func requireAuth() *entclient.Client {
-	client, err := newClient()
+func requireAuth(user string) *entclient.Client {
```
deansheather, line 14: How come you added the user to create the secret for to the struct? Wouldn't the API be better if it was just a param to the create secret function? Adding the user to the struct implies it's a sort of sudo thing like gitlab.

cmoog: My understanding is that we'll be expanding this concept to nearly all resources in the platform such that an admin could create/start/stop and environment for a specific user. So my thinking is that we'd want to parameterize the target user at the API client abstraction level.

deansheather: My issue with this however is that users of the API will find it tedious to have to recreate a new entclient every time if they wanted to do a Go-based solution for creating a secret for every user. You'd have to make an admin client for getting the user list, then a new client for every user.

How do other APIs handle sudo? I would've thought most APIs would do it like:
```go
type RequestOptions struct {
  Sudo string
}

var DefaultRequestOptions = &RequestOptions{
  Sudo: "me",
}

func (c *Client) GetSomething(ctx context.Context, opts *RequestOptions) {
  if (opts == nil) {
    opts = DefaultRequestOptions
  }

  ...
}
```

cmoog: Ah, this is a good point. Changing now.

PR #95, opened 2020-08-14
Cleanup and export entclient

cmoog:
https://github.com/cdr/enterprise/issues/4801


coder-sdk/env.go
```diff
@@ -33,12 +34,12 @@ type Environment struct {
 	AutoOffThreshold xjson.Duration `json:"auto_off_threshold" tab:"-"`
 }
 
-// Envs gets the list of environments owned by the authenticated user
-func (c Client) Envs(ctx context.Context, user *User, org Org) ([]Environment, error) {
+// EnvironmentsInOrganization gets the list of environments owned by the authenticated user
+func (c Client) EnvironmentsInOrganization(ctx context.Context, user *User, org *Org) ([]Environment, error) {
```
coadler, line 38: Is there a benefit to requiring the full `User` and `Org` structs here? Also, I think `EnvironmentsByOrganization` would be more consistent with other functions.

cmoog: Both valid points. just pushed the fixes.

PR #103, opened 2020-08-31
Remove dependency on HOME env var

### LINKED ISSUE #102: Windows distro uses HOME env var which is commonly not available
The HOME env var is not always set in a windows environments, causing config-ssh to fail not knowing where to put the private key.

cmoog:
Closes #102

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -52,11 +52,14 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		usr, err := user.Current()
+		if err != nil {
+			return xerrors.Errorf("get user home directory: %w", err)
+		}
+
+		privateKeyFilepath := filepath.Join(usr.HomeDir, ".ssh", "coder_enterprise")
+
 		if strings.HasPrefix(*configpath, "~") {
-			usr, err := user.Current()
-			if err != nil {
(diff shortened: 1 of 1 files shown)


wbobeirne:
Tagging in @creack as both a Go wizard, and a Windows user

cmoog:
> Tagging in @creack as both a Go wizard, and a Windows user

Sounds good. It should be noted that I am _definitely_ _not_ a windows user.

creack:
For the records, I am not either!, But I'll look into it :)

internal/cmd/configssh.go
```diff
@@ -52,11 +52,14 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		usr, err := user.Current()
```
creack, line 55: That is going to be a problem on some systems and will prevent us from shipping a static binaries.
It has been a while, I'm looking into this to see the extent, it might have improved.

creack: Alright, I didn't see we were using it before already.
I checked and at least in go1.15, it does work without cgo! Which is quite neat.

creack: Tested with and without cgo, and tested on windows, works as expected.
Slight note on windows, the path output has double `\`.
window ssh-config coder-cli

cmoog: Pushed a fix that resolves the double `\` (straight to master by accident 😰 ).

PR #110, opened 2020-09-02
Improve shell disconnect error

### LINKED ISSUE #87: coder sh prints usage when environment is offline
```
[willo@mbp ~]$ coder sh dev
2020-08-03 08:30:58 FATAL	run command: read pid message: failed to get reader: received close frame: status = StatusInternalError and reason = ""
```
We should have a better error message for common cases such as an off environment, a rebuilding environment etc.

cmoog:
Closes #87 

**Examples**

```
network error, environment status is "ON", please try again
network error, environment "dev" has a status of "OFF"
network error, environment "dev" has a status of "CREATING"
network error, environment "dev" has a status of "FAILED"
network error, environment "dev" has a status of "UNKNOWN"
```

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"strings"
@@ -142,11 +143,7 @@ func runCommand(ctx context.Context, envName string, command string, args []stri
 		Env:     cmdEnv,
 	})
 	if err != nil {
-		var closeErr websocket.CloseError
-		if xerrors.As(err, &closeErr) {
-			return xerrors.Errorf("network error, is %q online?", envName)
-		}
-		return err
+		return prettyShellExitError(ctx, err, entClient, env.Name)
 	}
 
 	if tty {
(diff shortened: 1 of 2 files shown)


internal/cmd/shell.go
```diff
@@ -178,15 +175,36 @@ func runCommand(ctx context.Context, envName string, command string, args []stri
...
-		return err
+		return prettyShellExitError(ctx, err, entClient, env.Name)
 	}
 	return nil
 }
 
+func prettyShellExitError(ctx context.Context, err error, client *coder.Client, envName string) error {
+	var closeErr websocket.CloseError
+	if xerrors.Is(err, ctx.Err()) || xerrors.As(err, &closeErr) {
+		if err != nil {
```
creack, line 186: we already check that the error is not nil before calling the function., no need to check again.
Would be better to add a comment on the function stating that the error passed to it is expected to be set, makes things easier to follow and avoid superfluous checks by simplifying the code.

internal/cmd/shell.go
```diff
@@ -178,15 +175,36 @@ func runCommand(ctx context.Context, envName string, command string, args []stri
...
+			return xerrors.Errorf("network error, %s", envErrorMessage(ctx, client, envName))
+		}
+	}
+	return err
+}
+
+func envErrorMessage(ctx context.Context, client *coder.Client, envName string) string {
+	env, err := client.EnvironmentByName(ctx, envName)
+	if err != nil {
+		return fmt.Sprintf("failed to fetch environment status: %v", err)
```
creack, line 196: here, we know the error is set, would be better with %s.

cmoog:
Closing this as it seems to have gotten stale and the error behavior has improved considerably.

PR #113, opened 2020-09-04
Minor clean / tidying up the code

creack:
- Scope errors when possible.
- Fix nil pointers in unmarshal.
- Align struct tags for readability.
- Explicitly discard errors with a comment on why.
- Import package order.
- Fmt verb for types.
- More comments.

### DIFF
--- a/ci/integration/setup_test.go
+++ b/ci/integration/setup_test.go
@@ -32,7 +32,7 @@ func init() {
 
 // build the coder-cli binary and move to the integration testing bin directory
 func build(path string) error {
-	tar := fmt.Sprintf("coder-cli-linux-amd64.tar.gz")
+	tar := "coder-cli-linux-amd64.tar.gz"
 	dir := filepath.Dir(path)
 	cmd := exec.Command(
 		"sh", "-c",
--- a/cmd/coder/main.go
+++ b/cmd/coder/main.go
@@ -15,14 +15,14 @@ import (
 	"go.coder.com/flog"
 )
 
-var (
-	version string = "unknown"
-)
+// Using a global for the version so it can be set at build time using ldflags.
+var version = "unknown"
 
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// If requested, spin up the pprof webserver.
 	if os.Getenv("PPROF") != "" {
 		go func() {
 			log.Println(http.ListenAndServe("localhost:6060", nil))
@@ -31,15 +31,19 @@ func main() {
 
 	stdoutState, err := xterminal.MakeOutputRaw(os.Stdout.Fd())
 	if err != nil {
-		flog.Fatal("set output to raw: %v", err)
+		flog.Fatal("set output to raw: %s", err)
 	}
-	defer xterminal.Restore(os.Stdout.Fd(), stdoutState)
+	defer func() {
+		// Best effort. Would result in broken terminal on window but nothing we can do about it.
+		_ = xterminal.Restore(os.Stdout.Fd(), stdoutState)
+	}()
 
 	app := cmd.Make()
 	app.Version = fmt.Sprintf("%s %s %s/%s", version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 
-	err = app.ExecuteContext(ctx)
-	if err != nil {
+	if err := app.ExecuteContext(ctx); err != nil {
+		// NOTE: The returned error is already handled and logged by the cmd lib (cobra), so no need to re-handle it here.
+		//       As we are in the main, if there was an error, exit the process with an error code.
 		os.Exit(1)
 	}
 }
--- a/coder-sdk/client.go
+++ b/coder-sdk/client.go
@@ -15,12 +15,17 @@ type Client struct {
 	Token   string
 }
 
-func (c Client) copyURL() *url.URL {
-	swp := *c.BaseURL
-	return &swp
-}
-
-func (c *Client) http() (*http.Client, error) {
+// newHTTPClient creates a default underlying http client and sets the auth cookie.
+//
+// NOTE: As we do not specify a custom transport, the default one from the stdlib will be used,
+//       resulting in a persistent connection pool.
+//       We do not set a timeout here as it could cause issue with the websocket.
+//       The caller is expected to set it when needed.
+//
+// WARNING: If the caller sets a custom transport to set TLS settings or a custom CA, the default
+//          pool will not be used and it might result in a new dns lookup/tls session/socket begin
+//          established each time.
+func (c *Client) newHTTPClient() (*http.Client, error) {
 	jar, err := cookiejar.New(nil)
 	if err != nil {
 		return nil, err
@@ -36,5 +41,6 @@ func (c *Client) http() (*http.Client, error) {
 			Secure:   c.BaseURL.Scheme == "https",
 		},
 	})
+
 	return &http.Client{Jar: jar}, nil
 }
--- a/coder-sdk/env.go
+++ b/coder-sdk/env.go
@@ -31,13 +31,18 @@ type Environment struct {
 	UpdatedAt        time.Time        `json:"updated_at"         tab:"-"`
 	LastOpenedAt     time.Time        `json:"last_opened_at"     tab:"-"`
 	LastConnectionAt time.Time        `json:"last_connection_at" tab:"-"`
-	AutoOffThreshold xjson.Duration   `json:"auto_off_threshold" tab:"-"`
+	AutoOffThreshold xjson.MSDuration `json:"auto_off_threshold" tab:"-"`
 }
 
 // RebuildMessage defines the message shown when an Environment requires a rebuild for it can be accessed.
 type RebuildMessage struct {
-	Text     string `json:"text"`
-	Required bool   `json:"required"`
+	Text             string           `json:"text"`
+	Required         bool             `json:"required"`
+	AutoOffThreshold xjson.MSDuration `json:"auto_off_threshold" tab:"-"`
+	RebuildMessages  []struct {
+		Text     string `json:"text"`
+		Required bool   `json:"required"`
+	} `json:"rebuild_messages" tab:"-"`
 }
 
 // EnvironmentStat represents the state of an environment
@@ -53,9 +58,7 @@ type EnvironmentStat struct {
 	DiskUsed        int64             `json:"disk_used"`
 }
 
-func (e EnvironmentStat) String() string {
-	return string(e.ContainerStatus)
-}
+func (e EnvironmentStat) String() string { return string(e.ContainerStatus) }
 
 // EnvironmentStatus refers to the states of an environment.
 type EnvironmentStatus string
@@ -69,7 +72,7 @@ const (
 	EnvironmentUnknown  EnvironmentStatus = "UNKNOWN"
 )
 
-// CreateEnvironmentRequest is used to configure a new environment
+// CreateEnvironmentRequest is used to configure a new environment.
 type CreateEnvironmentRequest struct {
 	Name     string   `json:"name"`
 	ImageID  string   `json:"image_id"`
@@ -84,61 +87,50 @@ type CreateEnvironmentRequest struct {
 // CreateEnvironment sends a request to create an environment.
 func (c Client) CreateEnvironment(ctx context.Context, orgID string, req CreateEnvironmentRequest) (*Environment, error) {
 	var env Environment
-	err := c.requestBody(
-		ctx,
-		http.MethodPost, "/api/orgs/"+orgID+"/environments",
-		req,
-		&env,
-	)
-	return &env, err
+	if err := c.requestBody(ctx, http.MethodPost, "/api/orgs/"+orgID+"/environments", req, &env); err != nil {
+		return nil, err
+	}
+	return &env, nil
 }
 
 // EnvironmentsByOrganization gets the list of environments owned by the given user.
 func (c Client) EnvironmentsByOrganization(ctx context.Context, userID, orgID string) ([]Environment, error) {
 	var envs []Environment
-	err := c.requestBody(
-		ctx,
-		http.MethodGet, "/api/orgs/"+orgID+"/members/"+userID+"/environments",
-		nil,
-		&envs,
-	)
-	return envs, err
+	if err := c.requestBody(ctx, http.MethodGet, "/api/orgs/"+orgID+"/members/"+userID+"/environments", nil, &envs); err != nil {
+		return nil, err
+	}
+	return envs, nil
 }
 
 // DeleteEnvironment deletes the environment.
 func (c Client) DeleteEnvironment(ctx context.Context, envID string) error {
-	return c.requestBody(
-		ctx,
-		http.MethodDelete, "/api/environments/"+envID,
-		nil,
-		nil,
-	)
+	return c.requestBody(ctx, http.MethodDelete, "/api/environments/"+envID, nil, nil)
 }
 
 // DialWsep dials an environments command execution interface
-// See github.com/cdr/wsep for details
+// See https://github.com/cdr/wsep for details.
 func (c Client) DialWsep(ctx context.Context, env *Environment) (*websocket.Conn, error) {
-	return c.dialWs(ctx, "/proxy/environments/"+env.ID+"/wsep")
+	return c.dialWebsocket(ctx, "/proxy/environments/"+env.ID+"/wsep")
 }
 
 // DialIDEStatus opens a websocket connection for cpu load metrics on the environment
 func (c Client) DialIDEStatus(ctx context.Context, envID string) (*websocket.Conn, error) {
-	return c.dialWs(ctx, "/proxy/environments/"+envID+"/ide/api/status")
+	return c.dialWebsocket(ctx, "/proxy/environments/"+envID+"/ide/api/status")
 }
 
-// DialEnvironmentBuildLog opens a websocket connection for the environment build log messages
+// DialEnvironmentBuildLog opens a websocket connection for the environment build log messages.
 func (c Client) DialEnvironmentBuildLog(ctx context.Context, envID string) (*websocket.Conn, error) {
-	return c.dialWs(ctx, "/api/environments/"+envID+"/watch-update")
+	return c.dialWebsocket(ctx, "/api/environments/"+envID+"/watch-update")
 }
 
-// DialEnvironmentStats opens a websocket connection for environment stats
+// DialEnvironmentStats opens a websocket connection for environment stats.
 func (c Client) DialEnvironmentStats(ctx context.Context, envID string) (*websocket.Conn, error) {
-	return c.dialWs(ctx, "/api/environments/"+envID+"/watch-stats")
+	return c.dialWebsocket(ctx, "/api/environments/"+envID+"/watch-stats")
 }
 
 // DialResourceLoad opens a websocket connection for cpu load metrics on the environment
 func (c Client) DialResourceLoad(ctx context.Context, envID string) (*websocket.Conn, error) {
-	return c.dialWs(ctx, "/api/environments/"+envID+"/watch-resource-load")
+	return c.dialWebsocket(ctx, "/api/environments/"+envID+"/watch-resource-load")
 }
 
 // BuildLogType describes the type of an event.
--- a/coder-sdk/error.go
+++ b/coder-sdk/error.go
@@ -11,6 +11,7 @@ import (
 // ErrNotFound describes an error case in which the requested resource could not be found
 var ErrNotFound = xerrors.Errorf("resource not found")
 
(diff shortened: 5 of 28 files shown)


coder-sdk/secrets.go
```diff
@@ -19,26 +19,36 @@ type Secret struct {
...
 // SecretWithValueByName gets the Coder secret with its value by its name.
 func (c *Client) SecretWithValueByName(ctx context.Context, name, userID string) (*Secret, error) {
+	// Lookup the secret from the name.
 	s, err := c.SecretByName(ctx, name, userID)
 	if err != nil {
 		return nil, err
 	}
+	// Pull the secret value.
+	// NOTE: This is racy, but acceptable. If the secret is gone or the permission changed since we looked up the id,
+	//       the call will simply fail and surface the error to the user.
```
cmoog, line 37: I don't see this as particularly actionable... if the secret is gone then the CLI will surface a 404 as expected. Changes in permissions would also result in the expected error.

internal/cmd/auth.go
```diff
@@ -3,42 +3,41 @@ package cmd
...
 	"go.coder.com/flog"
 )
 
-// requireAuth exits the process with a nonzero exit code if the user is not authenticated to make requests
+// requireAuth exits the process with a nonzero exit code if the user is not authenticated to make requests.
 func requireAuth() *coder.Client {
 	client, err := newClient()
 	if err != nil {
-		flog.Fatal("%v", err)
+		flog.Fatal("%s", err)
```
cmoog, line 18: I've been meaning to remove this in favor of returning a proper error.

internal/cmd/shell.go
```diff
@@ -196,9 +203,10 @@ func heartbeat(ctx context.Context, c *websocket.Conn, interval time.Duration) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			err := c.Ping(ctx)
-			if err != nil {
-				flog.Fatal("\nFailed to ping websocket: %v, exiting...", err)
+			if err := conn.Ping(ctx); err != nil {
+				// TODO: Don't fatal, it would skip the TTY state restore and result in a broken terminal.
```
cmoog, line 207: Ah! Nice catch.

cmoog (approved):
Thanks 😄

internal/x/xterminal/terminal_windows.go
```diff
@@ -85,29 +88,38 @@ func (s ResizeEvent) equal(s2 *ResizeEvent) bool {
...
+		var lastEvent *ResizeEvent
 		for {
 			select {
 			case <-ctx.Done():
 				return
 			case <-ticker.C:
-				width, height, err := terminal.GetSize(int(windows.Handle(termfd)))
+				width, height, err := terminal.GetSize(int(windows.Handle(termFD)))
 				if err != nil {
+					// TODO: Consider doing a debug log and continue instead of sopping the routine.
```
coadler, line 110:
```suggestion
					// TODO: Consider doing a debug log and continue instead of stopping the routine.
```

coadler (approved):
Just some random typos I found. Great improvements :smiley:

ci/integration/setup_test.go
```diff
@@ -32,7 +32,7 @@ func init() {
 
 // build the coder-cli binary and move to the integration testing bin directory
 func build(path string) error {
-	tar := fmt.Sprintf("coder-cli-linux-amd64.tar.gz")
```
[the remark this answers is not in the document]
cmoog: Yeah, that one is particularly embarrassing... leftover from when it was MacOS and Linux. 😭

cmd/coder/main.go
```diff
@@ -15,31 +15,35 @@ import (
...
+// Using a global for the version so it can be set at build time using ldflags.
+var version string = "unknown"
 
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// If requested, spin up the pprof webserver.
 	if os.Getenv("PPROF") != "" {
 		go func() {
```
tychoish, line 27: I wish we put panic handlers in all go routines.

cmd/coder/main.go
```diff
@@ -15,31 +15,35 @@ import (
...
 			log.Println(http.ListenAndServe("localhost:6060", nil))
 		}()
 	}
 
+	// TODO: Document why this is needed. Only does something on windows.
```
tychoish, line 32: I'd be inclined to avoid adding TODOs to the code base? I think it's better to either do the thing if its small or open issues?

cmd/coder/main.go
```diff
@@ -15,31 +15,35 @@ import (
...
+		_ = xterminal.Restore(os.Stdout.Fd(), stdoutState)
+	}()
 
 	app := cmd.Make()
 	app.Version = fmt.Sprintf("%s %s %s/%s", version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 
-	err = app.ExecuteContext(ctx)
-	if err != nil {
+	if err := app.ExecuteContext(ctx); err != nil {
+		// NOTE: The cmd lib will handle the logs, no need to do it here as well, simply exit with error code.
```
tychoish, line 46: what about this error? is it interesting?

coder-sdk/client.go
```diff
@@ -36,5 +32,7 @@ func (c *Client) http() (*http.Client, error) {
 			Secure:   c.BaseURL.Scheme == "https",
 		},
 	})
+	// TODO: Consider setting a timeout.
+	// TODO: Consider having a persistent client to allow for single connection.
```
tychoish, line 36: add bug reports for this? I think timeouts are great. 

I'd also be inclined in doing something to pool/reuse clients (e.g. https://github.com/deciduosity/utility/blob/main/http.go)

coder-sdk/env.go
```diff
@@ -81,50 +84,39 @@ type CreateEnvironmentRequest struct {
...
-	var env *Environment
-	err := c.requestBody(
-		ctx,
-		http.MethodPost, "/api/orgs/"+orgID+"/environments",
-		req,
-		env,
-	)
-	return env, err
+	var env Environment
+	if err := c.requestBody(ctx, http.MethodPost, "/api/orgs/"+orgID+"/environments", req, &env); err != nil {
```
tychoish, line 88: sprintf rather than string concatenation?

coder-sdk/error.go
```diff
@@ -2,18 +2,20 @@ package coder
...
 	"net/http"
 	"net/http/httputil"
 
 	"golang.org/x/xerrors"
 )
 
-// ErrNotFound describes an error case in which the requested resource could not be found
-var ErrNotFound = xerrors.Errorf("resource not found")
+// ErrNotFound describes an error case in which the requested resource could not be found.
+var ErrNotFound = errors.New("resource not found")
```
tychoish, line 13: I thought we were pro-xerrors?

coder-sdk/env.go
```diff
@@ -81,50 +84,39 @@ type CreateEnvironmentRequest struct {
...
 	var envs []Environment
-	err := c.requestBody(
-		ctx,
-		http.MethodGet, "/api/orgs/"+orgID+"/members/"+userID+"/environments",
-		nil,
-		&envs,
-	)
-	return envs, err
+	if err := c.requestBody(ctx, http.MethodGet, "/api/orgs/"+orgID+"/members/"+userID+"/environments", nil, &envs); err != nil {
+		return nil, err
```
tychoish, line 98: should we be wrapping these as discussed?

coder-sdk/request.go
```diff
@@ -4,51 +4,62 @@ import (
...
-	req, err := http.NewRequestWithContext(ctx, method, c.BaseURL.String()+path, bytes.NewReader(body))
+
+	// Create the http request.
+	req, err := http.NewRequestWithContext(ctx, method, c.BaseURL.String()+path, payload)
 	if err != nil {
 		return nil, xerrors.Errorf("create request: %w", err)
 	}
+	// TODO: Consider setting content-type / encoding / accept headers.
+
+	// Execute the request.
```
tychoish, line 38: I don't think it's necessary to have comments to describe function calls.

also, omit the todo or opening a ticket?

internal/activity/pusher.go
```diff
@@ -4,13 +4,14 @@ import (
...
-	"cdr.dev/coder-cli/coder-sdk"
 	"golang.org/x/time/rate"
 
+	"cdr.dev/coder-cli/coder-sdk"
+
 	"go.coder.com/flog"
 )
 
-const pushInterval = time.Minute
+const pushInterval = 1 * time.Minute
```
tychoish, line 14: I wouldn't do this.

internal/cmd/cmd.go
```diff
@@ -92,13 +91,13 @@ $ coder completion fish > ~/.config/fish/completions/coder.fish
 	Run: func(cmd *cobra.Command, args []string) {
 		switch args[0] {
 		case "bash":
-			cmd.Root().GenBashCompletion(os.Stdout)
+			_ = cmd.Root().GenBashCompletion(os.Stdout) // Best effort.
```
tychoish, line 94: I would only make the best effort comment once?

internal/cmd/shell.go
```diff
@@ -95,32 +96,39 @@ func sendResizeEvents(ctx context.Context, termfd uintptr, process wsep.Process)
...
 	if err != nil {
-		return err
+		return xerrors.Errorf("find environment: %w", err)
 	}
 
-	termfd := os.Stdout.Fd()
+	termFD := os.Stdout.Fd()
 
-	tty := terminal.IsTerminal(int(termfd))
+	// Check if the client is running an interactive session.
```
tychoish, line 109: this seems redundant to the code. 

I'm generally hesitant on adding comments that are redundant to the code in cases like this where the comments just describe what the code does, which just can get out of date.

internal/cmd/shell.go
```diff
@@ -196,9 +203,10 @@ func heartbeat(ctx context.Context, c *websocket.Conn, interval time.Duration) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			err := c.Ping(ctx)
-			if err != nil {
-				flog.Fatal("\nFailed to ping websocket: %v, exiting...", err)
+			if err := conn.Ping(ctx); err != nil {
+				// TODO: Don't fatal, it would skip the TTY state restore and result in a broken terminal.
+				// NOTE: Prefix with \r\n to attempt to have clearer output as we might still be in raw mode.
```
tychoish, line 208: can the prefix be a constant that we can reuse for these kinds of situations?

internal/sync/sync.go
```diff
@@ -91,74 +91,70 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
...
+	go func() { _, _ = io.Copy(os.Stderr, process.Stderr()) }() // Best effort.
 
-	err = process.Wait()
-	if code, ok := err.(wsep.ExitError); ok {
-		return fmt.Errorf("%s exit status: %v", prog, code)
-	}
-	if err != nil {
+	if err = process.Wait(); err != nil {
+		if code, ok := err.(wsep.ExitError); ok {
+			return fmt.Errorf("%s exit status: %d", prog, code)
```
tychoish, line 112: we should be consistent, I think about which Errorf questions.

internal/sync/sync.go
```diff
@@ -91,74 +91,70 @@ func (s Sync) syncPaths(delete bool, local, remote string) error {
...
 		}
-
 		return err
 	}
 	return nil
 }
 
 func (s Sync) handleDelete(localPath string) error {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
tychoish, line 157: I wouldn't multiply by 1?

internal/sync/sync.go
```diff
@@ -215,7 +211,7 @@ var ErrRestartSync = errors.New("the sync exited because it was overloaded, rest
 
 // workEventGroup converges a group of events to prevent duplicate work.
 func (s Sync) workEventGroup(evs []timedEvent) {
-	cache := make(eventCache)
+	cache := make(eventCache, len(evs))
```
tychoish, line 214: what Add do? this seems suspicious?

internal/x/xterminal/terminal_windows.go
```diff
@@ -85,29 +88,38 @@ func (s ResizeEvent) equal(s2 *ResizeEvent) bool {
 // ResizeEvents sends terminal resize events when the dimensions change.
 // Windows does not have a unix.SIGWINCH equivalent, so we poll the terminal size
 // at a fixed interval
-func ResizeEvents(ctx context.Context, termfd uintptr) chan ResizeEvent {
-	events := make(chan ResizeEvent)
-	ticker := time.NewTicker(time.Millisecond * 100)
+func ResizeEvents(ctx context.Context, termFD uintptr) chan ResizeEvent {
+	events := make(chan ResizeEvent, 1) // Use a buffered chan to avoid blocking.
```
tychoish, line 92: isn't blocking a symptom of a bug? if it works without buffering, I would leave it buffered?

tychoish (changes requested):

coder-sdk/secrets.go
cmoog: Ah, I see. Thanks 👍 .

internal/cmd/auth.go
creack: https://github.com/cdr/coder-cli/issues/114

internal/x/xterminal/terminal_windows.go
creack: I'll update the comment to add more context about the reasoning.

We want to send an initial event to signal the current size of the window. I was planning to do this in a different PR as this one ended up being mostly about styling/cleaning.

At a higher level, the idea is that when dealing with the channel, the runtime notifies us via a non-blocking chan, so if we are consuming it on an unbuffered chan, we lose the event.
In this specific case, we are dealing with windows and it is not actually a signal, but it maps directly with how it works on unix.

internal/sync/sync.go
creack: We are creating a map where we know exactly how big it is going to be in advance, so we can pre-allocate the memory for it.

internal/sync/sync.go
creack: for consistency and readability, I think it is best to always specify the full time, it shows it is intended and not an oversight. As it is a const, this is handled at compile time and have no impact and the runtime, it is purely aesthetic.

With this in mind, if you think it is best to remove it, we sure can.

internal/sync/sync.go
creack: I was wondering about it, as it was already there, I opted not to change it as there might have been a reason?

internal/cmd/shell.go
creack: This is a one-off that should be avoided, as we should control how/what we output. Promoting this to a const would imply it is supported and might encourage it to be used and for just two chars, especially CRLF, I am not sure it would add value.

internal/cmd/shell.go
creack: I am not sure most people make the relative between IsTerminal and interactive. We are dealing with somewhat advanced low level programming, following the execution flow in plain engligh helps understanding.

Out of date comments is indeed a problem, but it is on us not to let that happen :)

coder-sdk/env.go
creack: There is a mix of both a bit everywhere in the repo. Here I just stopped the error without changing how the string is created.

Created #115 to track this.

cmd/coder/main.go
creack: I'll reword the comment to be clearer. That error is already checked and log in the cmd lib. It is indeed an interesting one :)

cmd/coder/main.go
creack: Will remove the todos in favor of issues.

cmd/coder/main.go
creack: Hmm, panics are developer errors, not really expected in production. As a safety net on a service to log/alert about a bug makes sense, and in general, for anything expected to be long-running, I do agree, we should have panic handler, but here, we are in a CLI and specifically in a debug handler, so it is probably not necessary.

coder-sdk/error.go
creack: From chatting with @nhooyr we are using xerrors for the >go1.13 formatting/warps but also for the stack trace. Here we don't have any wrap/formatting, and as we are dealing with a "sentinel" error, having a trace indicated error.gol13 doesn't provide value and can lead to confusion.

coder-sdk/env.go
creack: When there is only one error path, imho it is not necessary, but more is better, I'll add it.

internal/cmd/cmd.go
creack: I actually strongly disagree, it is about the concept, there is an error, it is discarded, we indicate it is on purpose, this should be done each and every time. Ideally with more details, but at the very least with a message showing the intent.

internal/activity/pusher.go
creack: Can you extend on why? imho, it makes things easier to read and shows that it is a meant like the actual interval and not a unit

coder-sdk/request.go
creack: I am not sure everyone makes the relation between `req.Do` and effectively executing the request, especially when having `http.NewRequest` earlier. Having comments to explain the flow in plain English makes things easier to follow and maintain.

If you think it is just too much, I"ll remove it.

coder-sdk/client.go
creack: Will do.

Using a 3rd party pooling here might be overkill, but it would be interesting in our services

coder-sdk/error.go
nhooyr: I think it's good to be consistent though. If we use `xerrors.New``, I think we're good.

coder-sdk/client.go
tychoish: The pool seems to only be needed when you're playing with multiple transports, I think it's fine as is for now.

coder-sdk/request.go
tychoish: I'm still for removing. 

Comments are just one more thing that can get crufty, and while I agree that the interface is intuitive, I would hate that we should sprinkle descriptions of other package's APIs throughout calling code.

internal/activity/pusher.go
tychoish: I always assume that people typo'd `10*time.Minute` here, and think that time constants are clear enough.

internal/activity/pusher.go
```diff
@@ -19,27 +20,28 @@ type Pusher struct {
 	source string
 
 	client *coder.Client
-	rate   *rate.Limiter
+	rate   *rate.Limiter // Use a rate limiter to control the sampling rate.
```
tychoish, line 23: I would omit this comment too.

internal/cmd/cmd.go
tychoish: I think putting a higher level comment that says "we ignore errors when generating completion because they're not critical and the tool degrades gracefully" is very clear, whereas "best effort" doesn't really capture that decision or background as clearly (particularly when redundant.) 

I think the ignored error itself does a pretty good job of communicating intent, though I agree that some kind of comment is useful.

I'm not strongly committed to this view.

internal/cmd/shell.go
tychoish: The best way to avoid having out of date comments is to avoid writing code that needs comments :) 

Perhaps the `tty` variable would be more clear if it was named `isInteractive` or something similar. It could also be combined onto the next line. 

I think given how internal/low level (I suppose), I think it's fine

internal/cmd/shell.go
tychoish: I think putting "LineFeed" as constant gives it a name which makes the comment less required?

I feel like adding `\r\n` to the ends of something is never just a one-off, but I'm far less familiar with this code base.

internal/sync/sync.go
tychoish: I'm not sure, I can't imagine a way that it would matter, and it feels like this is the right kind of PR to make that change. I don't feel particularly strongly at all.

internal/sync/sync.go
tychoish: I'm still for removing, in a "less is more" sort of approach.

internal/sync/sync.go
tychoish: Interesting, I guess not seeing what type `eventCache` is makes this harder. I'm all down for preallocation, but we're far enough from the context that it was hard to read/guess. 

Frankly, it'd make more sense if there was a constructor that did all of this, rather than just having it at the beginning of a worker pool.

tychoish (approved):
I don't think anything is blocking merging this, but we can talk a little bit more about any of the details. I think most of these changes are an improvement.

PR #124, opened 2020-09-07
Manual token input during login.

### LINKED ISSUE #107: Add input while waiting for oauth callback to allow ssh
Right now, the cli login expects a HTTP callback, so when working over ssh, need to manually copy/paste the link into a different terminal and curl it. Having a text input to do the copy/paste directly on the cli would help a lot. (That's how concourse handles it for instance).
To be clear, that would be in addition of what we have, not instead.

creack:
Loosely based on how concourse ci handles it.

Fixes #107. This makes for a better experience when using the cli over ssh.

Notable change: use the full URL given by the user instead of trimming down the path. This allows for path based reverse proxies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/rjeczalik/notify v0.9.2
 	github.com/spf13/cobra v1.0.0
-	github.com/stretchr/testify v1.6.1 // indirect
+	github.com/stretchr/testify v1.6.1
 	go.coder.com/flog v0.0.0-20190906214207-47dd47ea0512
 	golang.org/x/crypto v0.0.0-20200422194213-44a606286825
 	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -1,83 +1,149 @@
 package cmd
 
 import (
+	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
-	"sync"
 
+	"cdr.dev/coder-cli/coder-sdk"
 	"cdr.dev/coder-cli/internal/config"
 	"cdr.dev/coder-cli/internal/loginsrv"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/xerrors"
 
 	"go.coder.com/flog"
 )
 
 func makeLoginCmd() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "login [Coder Enterprise URL eg. http://my.coder.domain/]",
+	return &cobra.Command{
+		Use:   "login [Coder Enterprise URL eg. https://my.coder.domain/]",
 		Short: "Authenticate this client for future operations",
 		Args:  cobra.ExactArgs(1),
-		RunE:  login,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Pull the URL from the args and do some sanity check.
+			rawURL := args[0]
+			if rawURL == "" || !strings.HasPrefix(rawURL, "http") {
+				return xerrors.Errorf("invalid URL")
+			}
+			u, err := url.Parse(rawURL)
+			if err != nil {
+				return xerrors.Errorf("parse url: %w", err)
+			}
+			// Remove the trailing '/' if any.
+			u.Path = strings.TrimSuffix(u.Path, "/")
+
+			// From this point, the commandline is correct.
+			// Don't return errors as it would print the usage.
+
+			if err := login(cmd, u, config.URL, config.Session); err != nil {
+				flog.Error("Login error: %s.", err)
+				os.Exit(1)
+			}
+
+			return nil
+		},
 	}
-	return cmd
 }
 
-func login(cmd *cobra.Command, args []string) error {
-	rawURL := args[0]
-	if rawURL == "" || !strings.HasPrefix(rawURL, "http") {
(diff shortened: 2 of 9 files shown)


internal/cmd/login.go
```diff
@@ -1,83 +1,149 @@
...
+		return xerrors.Errorf("ping API: %w", err)
 	}
-	flog.Success("logged in")
+
+	// Success. Store the config only at this point so we don't override the local one in case of failure.
+	if err := storeConfig(envURL, token, urlCfg, sessionCfg); err != nil {
+		return xerrors.Errorf("store config: %w", err)
+	}
+
+	flog.Success("Logged in.")
```
cmoog, line 146: Should be precede this with a newline? Right now, the output is:
```
$ coder login https://master.cdr.dev
or enter token manually: 2020-09-08 16:04:30 SUCCESS    Logged in.
```

internal/loginsrv/input.go
```diff
@@ -0,0 +1,58 @@
...
+
+// ReadLine waits for the manual login input to send the session token.
+// NOTE: As we are dealing with a Read, cancelling the context will not unblock.
+//       The caller is expected to close the reader.
+func ReadLine(ctx context.Context, r io.Reader, w io.Writer, tokenChan chan<- string) error {
+	// Wrap the reader with bufio to simplify the readline.
+	buf := bufio.NewReader(r)
+
+retry:
+	_, _ = fmt.Fprintf(w, "or enter token manually: ") // Best effort. Can only fail on custom writers.
```
cmoog, line 22: This message doesn't seem descriptive enough... maybe we should include specific direction so the user knows to paste the contents of the URL bar.

cmoog (changes requested):
First pass. Good thinking with keeping the full URL in the config.

internal/cmd/login.go
```diff
@@ -1,83 +1,149 @@
...
+			}
+			u, err := url.Parse(rawURL)
+			if err != nil {
+				return xerrors.Errorf("parse url: %w", err)
+			}
+			// Remove the trailing '/' if any.
+			u.Path = strings.TrimSuffix(u.Path, "/")
+
+			// From this point, the commandline is correct.
+			// Don't return errors as it would print the usage.
```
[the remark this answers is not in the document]
creack: This seem a bit outside the scope of this PR as it is the way all commands are set. We can create a new PR to improve error management everywhere at once.

internal/loginsrv/input_test.go
```diff
@@ -0,0 +1,125 @@
...
+			case err := <-errChan:
+				t.Fatalf("ReadLine returned before we got the token (%v).", err)
+			case <-doneChan:
+			}
+
+			select {
+			case <-ctx.Done():
+				t.Fatal("Timeout waiting for the input.")
+			case err := <-errChan:
+				t.Fatalf("ReadLine returned before we got the token (%v).", err)
```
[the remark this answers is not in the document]
creack: I think we should stick with stdlib and standard widely used and supported instead of using custom ones. Unless there is some value I am missing? Integration with our co/monitoring/alerting maybe?

internal/loginsrv/input.go
creack: This is the same message that concourse ci uses, with the same formatting. We expect most people to not require this part, so I think it might be good enough? I am not really good with words ^^, If you have a suggestion for better wording, I'd be happy to update.

internal/loginsrv/input_test.go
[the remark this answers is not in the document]
coadler: I enjoy using slog since it's able to print the stack frames from the provided error, though I can't say there's much consistency in either direction atm.

internal/cmd/login.go
```diff
@@ -1,83 +1,149 @@
...
-	u, err := url.Parse(rawURL)
+// newLocalListener creates up a local tcp server using port 0 (i.e. any available port).
+// If ipv4 is disabled, try ipv6.
+// It will be used by the http server waiting for the auth callback.
+func newLocalListener() (net.Listener, error) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
 	if err != nil {
-		return xerrors.Errorf("parse url: %v", err)
+		if l, err = net.Listen("tcp6", "[::1]:0"); err != nil {
+			return nil, xerrors.Errorf("failed to listen on a port: %s", err)
```
coadler, line 61:
```suggestion
			return nil, xerrors.Errorf("listen on a port: %w", err)
```

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/rjeczalik/notify v0.9.2
 	github.com/spf13/cobra v1.0.0
-	github.com/stretchr/testify v1.6.1 // indirect
+	github.com/stretchr/testify v1.6.1
```
coadler, line 17: We've pretty much replaced all testify usage with https://pkg.go.dev/cdr.dev/slog@v1.3.0/sloggers/slogtest/assert

coadler: Main difference is `NoError` -> `Success` and the string msg always comes first.

coadler (approved):

go.mod
creack: There are some core issues with slog and I think we should try to move away from it.
testify is a widely used and well-supported lib, using it instead of a custom in-house one makes things easier to contribute for anyone within the company as well as from the community.

internal/cmd/login.go
creack: Added a newline in the prompt before this line. Otherwise, flog's prefix still is on he previous line.

cmoog (approved):
The language of `or enter token manually` is a bit off, but I don't have any better ideas at the moment. I'd be curious to hear what others think when they try it out after this lands.

PR #125, opened 2020-09-13
coder-sdk: remove spurious JSON tag

alrs:
The Go `json` struct tag knows nothing of the option `required`.

https://golang.org/pkg/encoding/json/


creack:
Thanks!

PR #126, opened 2020-09-15
Build for windows when building on linux

deansheather:
We weren't building the windows version on anything except darwin. Darwin will now build all versions, but linux will only build windows and linux versions.


cmoog (approved):
LGTM

PR #133, opened 2020-10-07
Give better error when authentication expires [ch134]

ammario:
## Before
image

## After

Screen Shot 2020-10-07 at 6 21 58 PM

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -71,7 +71,7 @@ func shell(_ *cobra.Command, cmdArgs []string) error {
 		if exitErr, ok := err.(wsep.ExitError); ok {
 			os.Exit(exitErr.Code)
 		}
-		return xerrors.Errorf("run command: %w", err)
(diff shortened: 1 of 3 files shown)


internal/cmd/shell.go
```diff
@@ -71,7 +71,7 @@ func shell(_ *cobra.Command, cmdArgs []string) error {
 		if exitErr, ok := err.(wsep.ExitError); ok {
 			os.Exit(exitErr.Code)
 		}
-		return xerrors.Errorf("run command: %w", err)
```
cmoog, line 74: Ideally we would be able to handle this better. This failure case _should_ propagate an error such that we exit with some nonzero code. I think there may be a way for us to hook into the cobra error handling so we can fix the ugly error formatting.

cmoog (approved):

cmoog (approved):
I'll fix the cobra formatting in a new PR so we can remove the scattered calls to flog.Fatal

PR #138, opened 2020-10-17
added quotes for identity to

KevinEdry:
I have added quotes for the identity file to enable it to work with paths that have white spaces (in a windows username for example).


cmoog:
@KevinEdry Nice catch 💯 . Thanks for the PR!

PR #139, opened 2020-10-17
fixed ssh key error to better reflect its state.

KevinEdry:
Fixed the errors coming from the ssh key write process to better reflect when the ssh key already exists or not.

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -130,7 +130,7 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 		}
 		err = writeSSHKey(ctx, client, privateKeyFilepath)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/cmd/configssh.go
```diff
@@ -130,11 +130,12 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 		}
 		err = writeSSHKey(ctx, client, privateKeyFilepath)
 		if err != nil {
-			return xerrors.Errorf("fetch and write ssh key: %w", err)
+			fmt.Printf("Your private ssh key already exists at \"%s\"\n", privateKeyFilepath)
```
cmoog, line 133:
```suggestion
			fmt.Printf("Your private ssh key already exists at \"%s\"\nYou may need to remove the existing file and re-run this command\n", privateKeyFilepath)
```

cmoog:
@KevinEdry Thank you for the contribution! 💯

PR #142, opened 2020-10-19
Add resource top flags for filtering and sorting

cmoog:
This PR implements the following flags for `coder resources top`

```
      --org string       filter by the name of an organization
      --show-empty       show groups with zero active environments
      --sort-by string   field to sort aggregate groups and environments by (cpu|memory) (default "cpu")
      --user string      filter by a user email
```


fuskovic (approved):
LGTM

PR #143, opened 2020-10-19
Add install script and readme snippet

cmoog:
Inspired by the simplicity of https://raw.githubusercontent.com/denoland/deno_install/master/install.sh

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ fi
 
 if [ "$OS" = "Windows_NT" ]; then
   target="windows-386"
-  ext=".zip"
+  extension=".zip"
   if ! command -v unzip >/dev/null; then
     echo "Error: unzip is required to install coder-cli" 1>&2
     exit 1
@@ -20,17 +20,18 @@ else
     echo "Error: tar is required to install coder-cli" 1>&2
     exit 1
   fi
-  ext=".tar.gz"
+  extension=".tar.gz"
   case $(uname -s) in
   Darwin) target="darwin-amd64" ;;
   *) target="linux-amd64" ;;
   esac
 fi
 
-if [ $# -eq 0 ]; then
+version=${1:-""}
+if [ "$version" == "" ]; then
   coder_asset_path=$(
     curl -sSf https://github.com/cdr/coder-cli/releases |
-      grep -o "/cdr/coder-cli/releases/download/.*/coder-cli-${target}${ext}" |
+      grep -o "/cdr/coder-cli/releases/download/.*/coder-cli-${target}${extension}" |
       head -n 1
   )
   if [ ! "$coder_asset_path" ]; then
@@ -39,7 +40,7 @@ if [ $# -eq 0 ]; then
   fi
   cdr_uri="https://github.com${coder_asset_path}"
 else
-  cdr_uri="https://github.com/cdr/coder-cli/releases/download/${1}/coder-cli-${target}${ext}"
+  cdr_uri="https://github.com/cdr/coder-cli/releases/download/${1}/coder-cli-${target}${extension}"
 fi
 
 coder_install="${CODER_INSTALL:-$HOME/.coder}"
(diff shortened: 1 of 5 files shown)


install.sh
```diff
@@ -0,0 +1,74 @@
...
+# TODO(everyone): Keep this script simple and easily auditable.
+
+set -e
+
+if [ "$(uname -m)" != "x86_64" ]; then
+  echo "Error: Unsupported architecture $(uname -m). Only x64 binaries are available." 1>&2
+  exit 1
+fi
+
+if [ "$OS" = "Windows_NT" ]; then
```
nhooyr, line 11: Does windows define `OS`?

install.sh
```diff
@@ -0,0 +1,74 @@
...
+  target="windows-386"
+  ext=".zip"
+  if ! command -v unzip >/dev/null; then
+    echo "Error: unzip is required to install coder-cli" 1>&2
+    exit 1
+  fi
+else
+  if ! command -v tar >/dev/null; then
```
nhooyr, line 19: Why is this required? Won't it just exit out with command not found?

install.sh
```diff
@@ -0,0 +1,74 @@
...
+    exit 1
+  fi
+  ext=".tar.gz"
+  case $(uname -s) in
+  Darwin) target="darwin-amd64" ;;
+  *) target="linux-amd64" ;;
+  esac
+fi
+
+if [ $# -eq 0 ]; then
```
nhooyr, line 30: Let's use a `$version` var for this to make it clear.

install.sh
```diff
@@ -0,0 +1,74 @@
...
+  if [ ! "$coder_asset_path" ]; then
+    echo "Error: Unable to find latest coder-cli release on GitHub." 1>&2
+    exit 1
+  fi
+  cdr_uri="https://github.com${coder_asset_path}"
+else
+  cdr_uri="https://github.com/cdr/coder-cli/releases/download/${1}/coder-cli-${target}${ext}"
+fi
+
+coder_install="${CODER_INSTALL:-$HOME/.coder}"
```
nhooyr, line 45: This should be `~/.local/lib/coder`.

See https://github.com/cdr/code-server/blob/11b51058355caa60d21f59b09996d127776c8c84/install.sh#L42-L46

install.sh
```diff
@@ -0,0 +1,74 @@
...
+bin_dir="$coder_install/bin"
+exe="$bin_dir/coder"
+
+if [ ! -d "$bin_dir" ]; then
+  mkdir -p "$bin_dir"
+fi
+
+curl --fail --location --progress-bar --output "$exe$ext" "$cdr_uri"
```
nhooyr, line 53: Can we make the flag short here too?

nhooyr, line 53: `exe` and `ext` are very confusing side by side

nhooyr (changes requested):

install.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/sh
+# TODO(everyone): Keep this script simple and easily auditable.
```
nhooyr, line 2: I like this!

install.sh
cmoog: Much cleaner to handle gracefully with a specific error message.

install.sh
cmoog: Why though? Not sure why we'd intentionally make it harder to read 😆

install.sh
cmoog: Not quite sure, although I suspect the Deno authors have tested this properly. Either way, I'll put together a Github actions test for this that runs on windows.

install.sh
cmoog: Is there a specific standard for this that I'm unaware of? I find having these kinds of hidden directories in my home folder convenient. Especially if we add configuration files to this directory down the line.

install.sh
nhooyr: How is it cleaner? It bloats the install script when the shell itself already handles it thanks to `set -e`.

install.sh
nhooyr: Yes, the XDG standard. It's `~/.local`. And then `~/.config` for the configuration files.

install.sh
nhooyr: No one uses the long flags for `curl`, everyone knows the short flags so they're easier to read.

wbobeirne: I do not know the long flags and I appreciate that moogster went long.

vapurrmaid: Same here, I pretty much always use long flags wherever possible

nhooyr: Then let's be consistent with the above invocation of curl.

install.sh
cmoog: Just verified this.

install.sh
cmoog: So there are zero programs on my machine in `~/.local/lib`. Inclined to ignore the standard.

cmoog:
Opting for package managers instead.

PR #146, opened 2020-10-20
Add coder envs rebuild and watch-build commands

cmoog:
```
$ coder envs rebuild entdev --follow
Rebuild environment "entdev"? (will destroy any work outside of /home): y█
✅ -- 2020-10-20T00:43:06Z Ensuring private SSH key
✅ -- 2020-10-20T00:43:06Z Deleting old environment
✅ -- 2020-10-20T00:43:50Z Deleting old network isolation policy
✅ -- 2020-10-20T00:43:50Z Deleting old service
✅ -- 2020-10-20T00:43:51Z Getting default user from image
✅ -- 2020-10-20T00:43:56Z Ensuring environment has a home volume
✅ -- 2020-10-20T00:43:56Z Creating network isolation policy
✅ -- 2020-10-20T00:43:56Z Ensuring environment service
```

### DIFF
--- a/coder-sdk/env.go
+++ b/coder-sdk/env.go
@@ -122,6 +122,29 @@ func (c Client) StopEnvironment(ctx context.Context, envID string) error {
 	return c.requestBody(ctx, http.MethodPut, "/api/environments/"+envID+"/stop", nil, nil)
 }
 
+// UpdateEnvironmentReq defines the update operation, only setting
+// nil-fields.
+type UpdateEnvironmentReq struct {
+	ImageID              *string   `json:"image_id"`
+	ImageTag             *string   `json:"image_tag"`
+	CPUCores             *float32  `json:"cpu_cores"`
+	MemoryGB             *float32  `json:"memory_gb"`
+	DiskGB               *int      `json:"disk_gb"`
+	GPUs                 *int      `json:"gpus"`
+	Services             *[]string `json:"services"`
+	CodeServerReleaseURL *string   `json:"code_server_release_url"`
+}
+
+// RebuildEnvironment requests that the given envID is rebuilt with no changes to its specification.
+func (c Client) RebuildEnvironment(ctx context.Context, envID string) error {
+	return c.requestBody(ctx, http.MethodPatch, "/api/environments/"+envID, UpdateEnvironmentReq{}, nil)
+}
+
+// EditEnvironment modifies the environment specification and initiates a rebuild.
+func (c Client) EditEnvironment(ctx context.Context, envID string, req UpdateEnvironmentReq) error {
+	return c.requestBody(ctx, http.MethodPatch, "/api/environments/"+envID, req, nil)
+}
+
 // DialWsep dials an environments command execution interface
 // See https://github.com/cdr/wsep for details.
 func (c Client) DialWsep(ctx context.Context, env *Environment) (*websocket.Conn, error) {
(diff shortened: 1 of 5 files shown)


coder-sdk/env.go
```diff
@@ -138,6 +161,49 @@ func (c Client) DialEnvironmentBuildLog(ctx context.Context, envID string) (*web
...
+// errors that may occur when follow or parsing.
+type BuildLogFollowMsg struct {
+	BuildLog
+	Err error
+}
+
+// FollowEnvironmentBuildLog trails the build log of a Coder environment.
+func (c Client) FollowEnvironmentBuildLog(ctx context.Context, envID string) (<-chan BuildLogFollowMsg, error) {
+	ch := make(chan BuildLogFollowMsg)
+	ws, err := c.DialEnvironmentBuildLog(ctx, envID)
```
cmoog, line 185: @f0ssel @coadler Any thoughts on this pattern? Seems better to me than having a `BuildLogWatcher` type with methods for iterating when channels are designed for that.

internal/cmd/rebuild.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+	var s *spinner.Spinner
+	for l := range logs {
+		if l.Err != nil {
+			return l.Err
+		}
+		switch l.BuildLog.Type {
+		case coder.BuildLogTypeStart:
```
fuskovic, line 84: Do we need to do anything for this case?

internal/cmd/rebuild.go
```diff
@@ -0,0 +1,137 @@
...
+			s.Start()
+		case coder.BuildLogTypeSubstage:
+			// TODO(@cmoog) add verbose substage printing
+		case coder.BuildLogTypeError:
+			s.FinalMSG = fmt.Sprintf("%s %s", failure, strings.TrimPrefix(s.Suffix, "  "))
+			s.Stop()
+			fmt.Print(color.RedString("\t%s", l.BuildLog.Msg))
+			s = newSpinner()
+		case coder.BuildLogTypeDone:
+			s.Stop()
```
fuskovic, line 103: Is it possible that `s` could be nil here or is it safe to assume we'll always have assigned the return value of `newSpinner` to `s` before we get here?

internal/cmd/rebuild.go
```diff
@@ -0,0 +1,137 @@
...
+		Short:   "trail the build log of a Coder environment",
+		Args:    cobra.ExactArgs(1),
+		Hidden:  true, // TODO(@cmoog) un-hide
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			client, err := newClient()
+			if err != nil {
+				return err
+			}
+			env, err := findEnv(ctx, client, args[0], coder.Me)
```
fuskovic, line 125: Could we get an out of bounds panic here if `args` == nil ?

Edit: I looked at the cobra godoc and saw how our use of `cobra.ExactArgs(1)` on L117 protects us from that.

fuskovic (approved):

internal/cmd/rebuild.go
cmoog: Ah good catch. Assuming the build log records are properly formed, this *shouldn't* be an issue. That said, I'll push a change that better handles the nil case to avoid panics.

internal/cmd/rebuild.go
cmoog: The FE uses this to reset the UI logs, but since we're just trailing I don't think it makes sense to do anything. Plus we exit after the Done message is sent anyway.

fuskovic: That makes sense.

I think that explanation should be plugged in as a comment.

PR #149, opened 2020-10-21
Add envs create command

fuskovic:
# Usage
Adds the `create` command to the `envs` command.

```
Create a new environment under the active user.

Usage:
  coder envs create [environment_name] [flags]

Examples:
# create a new environment using default resource amounts
coder envs create --image id-of-imported-image my-env-name

# create a new environment using custom resource amounts
coder envs create --cores 4 --disk 100 --memory 8 --image id-of-imported-image --org id-of-existing-organization my-env-name

# using short-hand flags.
coder envs create -c 4 -d 100 -m 8 -i id-of-imported-image -o id-of-existing-organization my-env-name

Flags:
  -c, --cpu float32      number of cpu cores the environment should be provisioned with. (default 1)
  -d, --disk int         GB of disk storage an environment should be provisioned with. (default 10)
      --follow           follow buildlog after initiating rebuild
  -g, --gpus int         number GPUs an environment should be provisioned with.
  -h, --help             help for create
  -i, --image string     ID of the image to base the environment off of.
  -m, --memory float32   GB of RAM an environment should be provisioned with. (default 1)
  -o, --org string       ID of the organization the environment should be created under. (default "default")
  -t, --tag string       tag of the image the environment will be based off of. (default "latest")
```

# Example: 

### Create a simple environment with default resource values

default_resource_amounts

### Create a new environment with custom resource values. 

custom_resource_amounts

### DIFF
--- a/ci/steps/gendocs.sh
+++ b/ci/steps/gendocs.sh
@@ -11,12 +11,6 @@ rm -rf ./docs
 mkdir ./docs
 go run ./cmd/coder gen-docs ./docs
 
-# remove cobra footer from each file
-for filename in ./docs/*.md; do
-  trimmed=$(head -n -1 "$filename")
-  echo "$trimmed" >$filename
-done
-
 if [[ ${CI-} && $(git ls-files --other --modified --exclude-standard) ]]; then
   echo "Documentation needs generation:"
   git -c color.ui=always status | grep --color=no '\e\[31m'
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -14,6 +14,15 @@ import (
 	"golang.org/x/xerrors"
 )
 
+const (
+	defaultOrg              = "default"
+	defaultImgTag           = "latest"
+	defaultCPUCores float32 = 1
+	defaultMemGB    float32 = 1
+	defaultDiskGB           = 10
+	defaultGPUs             = 0
+)
+
 func envsCommand() *cobra.Command {
 	var outputFmt string
 	var user string
@@ -64,9 +73,9 @@ func envsCommand() *cobra.Command {
 	lsCmd.Flags().StringVarP(&outputFmt, "output", "o", "human", "human | json")
 	cmd.AddCommand(lsCmd)
 	cmd.AddCommand(stopEnvCommand(&user))
-
 	cmd.AddCommand(watchBuildLogCommand())
 	cmd.AddCommand(rebuildEnvCommand())
+	cmd.AddCommand(createEnvCommand())
 	return cmd
 }
 
@@ -124,4 +133,77 @@ coder envs --user charlie@coder.com ls -o json \
 			return nil
 		},
 	}
+}
+
+func createEnvCommand() *cobra.Command {
+	var (
+		org    string
+		img    string
+		tag    string
+		follow bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "create [environment_name]",
+		Short: "create a new environment.",
+		Args:  cobra.ExactArgs(1),
+		// Don't unhide this command until we can pass image names instead of image id's.
+		Hidden: true,
+		Long:   "Create a new environment under the active user.",
+		Example: `# create a new environment using default resource amounts
+coder envs create --image 5f443b16-30652892427b955601330fa5 my-env-name
+
+# create a new environment using custom resource amounts
+coder envs create --cpu 4 --disk 100 --memory 8 --image 5f443b16-30652892427b955601330fa5 my-env-name`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if img == "" {
+				return xerrors.New("image id unset")
+			}
+			// ExactArgs(1) ensures our name value can't panic on an out of bounds.
+			createReq := &coder.CreateEnvironmentRequest{
+				Name:     args[0],
+				ImageID:  img,
+				ImageTag: tag,
+			}
+			// We're explicitly ignoring errors for these because all of these flags
+			// have a non-zero-value default value set already.
+			createReq.CPUCores, _ = cmd.Flags().GetFloat32("cpu")
+			createReq.MemoryGB, _ = cmd.Flags().GetFloat32("memory")
+			createReq.DiskGB, _ = cmd.Flags().GetInt("disk")
+			createReq.GPUs, _ = cmd.Flags().GetInt("gpus")
+
+			client, err := newClient()
+			if err != nil {
+				return err
+			}
+
+			env, err := client.CreateEnvironment(cmd.Context(), org, *createReq)
+			if err != nil {
+				return xerrors.Errorf("create environment: %w", err)
+			}
+
+			clog.LogSuccess(
+				"creating environment...",
+				clog.BlankLine,
+				clog.Tip(`run "coder envs watch-build %q" to trail the build logs`, args[0]),
+			)
+
+			if follow {
+				if err := trailBuildLogs(cmd.Context(), client, env.ID); err != nil {
+					return err
+				}
+			}
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&org, "org", "o", defaultOrg, "ID of the organization the environment should be created under.")
+	cmd.Flags().StringVarP(&tag, "tag", "t", defaultImgTag, "tag of the image the environment will be based off of.")
+	cmd.Flags().Float32P("cpu", "c", defaultCPUCores, "number of cpu cores the environment should be provisioned with.")
+	cmd.Flags().Float32P("memory", "m", defaultMemGB, "GB of RAM an environment should be provisioned with.")
+	cmd.Flags().IntP("disk", "d", defaultDiskGB, "GB of disk storage an environment should be provisioned with.")
+	cmd.Flags().IntP("gpus", "g", defaultGPUs, "number GPUs an environment should be provisioned with.")
+	cmd.Flags().StringVarP(&img, "image", "i", "", "ID of the image to base the environment off of.")
+	cmd.Flags().BoolVar(&follow, "follow", false, "follow buildlog after initiating rebuild")
+	cmd.MarkFlagRequired("image")
+	return cmd
 }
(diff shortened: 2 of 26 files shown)


internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Create a new environment under the active user.",
+		Example: `
+# create a new environment using default resource amounts
+coder envs create \
+	--image id-of-imported-image \
+	my-env-name
+
+See Flags section for default values.
```
cmoog, line 150: I'd say we axe this.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+	my-env-name
+
+# same command this time using short-hand flags for a more succinct command experience.
+coder envs create \
+	-c 4 \
+	-d 100 \
+	-m 8 \
+	-i id-of-imported-image my-env-name \
+	-o id-of-existing-organization \
+	my-env-name
```
cmoog, line 168: This also feels unnecessary and makes the help section far larger than it needs to be.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+			if err != nil {
+				return xerrors.Errorf("create environment: %w", err)
+			}
+			flog.Success("Successfully created environment %q", env.Name)
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&org, "org", "o", defaultOrg, "ID of the organization the environment should be created under.")
+	cmd.Flags().StringVarP(&tag, "tag", "t", defaultImgTag, "The particular tag of the image the environment will be based off of.")
+	cmd.Flags().Float32P("cores", "c", defaultCPUCores, "The number of cpu cores the environment should be provisioned with.")
```
cmoog, line 202: `cores` feels more like the "unit". Would be like if the flag for memory was `GB` or something. `--cpu` feels more natural.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+				Name:     args[0],
+				ImageID:  img,
+				ImageTag: tag,
+			}
+			// We're explicitly ignoring errors for these because all of these flags
+			// have a non-zero-value default value set already.
+			createReq.CPUCores, _ = cmd.Flags().GetFloat32("cores")
+			createReq.MemoryGB, _ = cmd.Flags().GetFloat32("memory")
+			createReq.DiskGB, _ = cmd.Flags().GetInt("disk")
+			createReq.GPUs, _ = cmd.Flags().GetInt("gpus")
```
cmoog, line 185: Won't those defaults be overwritten?

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+func createEnvCommand() *cobra.Command {
+	var (
+		org      string
+		img      string
+		tag      string
+		services []string
+	)
+
+	cmd := &cobra.Command{
+		Use:    "create [flags] [environment-name]",
```
cmoog, line 139: Isn't `create [environment_name] [...flags]` a bit more idiomatic?

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+			client, err := newClient()
+			if err != nil {
+				return err
+			}
+
+			env, err := client.CreateEnvironment(cmd.Context(), org, *createReq)
+			if err != nil {
+				return xerrors.Errorf("create environment: %w", err)
+			}
+			flog.Success("Successfully created environment %q", env.Name)
```
cmoog, line 196: This message is a bit confusing because we haven't actually successfully created the environment _yet_. Let's add a flag for `--follow` that follows the build log. If the flag isn't specified, we should log a message that creates the correct expectations.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+		},
+	}
+	cmd.Flags().StringVarP(&org, "org", "o", defaultOrg, "ID of the organization the environment should be created under.")
+	cmd.Flags().StringVarP(&tag, "tag", "t", defaultImgTag, "The particular tag of the image the environment will be based off of.")
+	cmd.Flags().Float32P("cores", "c", defaultCPUCores, "The number of cpu cores the environment should be provisioned with.")
+	cmd.Flags().Float32P("memory", "m", defaultMemGB, "GB of RAM an environment should be provisioned with.")
+	cmd.Flags().IntP("disk", "d", defaultDiskGB, "GB of disk storage an environment should be provisioned with.")
+	cmd.Flags().IntP("gpus", "g", defaultGPUs, "The number GPUs an environment should be provisioned with.")
+	cmd.Flags().StringSliceP("services", "s", services, "The services that the environment should be initialized with.")
+	cmd.Flags().StringVarP(&img, "image", "i", "", "ID of the image to base the environment off of.")
```
cmoog, line 207: I'm not a huge fan of us requiring random unique identifiers as arguments. The optimal solution here feels like the following:
```
--image-id <id of image>
--image <name of image>
```

Where we error if both are set.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&org, "org", "o", defaultOrg, "ID of the organization the environment should be created under.")
+	cmd.Flags().StringVarP(&tag, "tag", "t", defaultImgTag, "The particular tag of the image the environment will be based off of.")
+	cmd.Flags().Float32P("cores", "c", defaultCPUCores, "The number of cpu cores the environment should be provisioned with.")
+	cmd.Flags().Float32P("memory", "m", defaultMemGB, "GB of RAM an environment should be provisioned with.")
+	cmd.Flags().IntP("disk", "d", defaultDiskGB, "GB of disk storage an environment should be provisioned with.")
+	cmd.Flags().IntP("gpus", "g", defaultGPUs, "The number GPUs an environment should be provisioned with.")
+	cmd.Flags().StringSliceP("services", "s", services, "The services that the environment should be initialized with.")
```
cmoog, line 206: Same issue here as with the `--image` flag. Honestly since we don't seem to be investing in `services`, it might be a good idea to axe this flag altogether for now.

internal/cmd/envs.go
```diff
@@ -117,3 +126,85 @@ coder envs --user charlie@coder.com ls -o json \
...
+	)
+
+	cmd := &cobra.Command{
+		Use:    "create [flags] [environment-name]",
+		Short:  "create a new environment.",
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Create a new environment under the active user.",
+		Example: `
+# create a new environment using default resource amounts
```
cmoog, line 145: Don't need the leading newline char.

cmoog (changes requested):

internal/cmd/envs.go
cmoog: Ah my bad. Didn't realize how you were using them.

internal/cmd/envs.go
fuskovic: The image-id is a hard requirement for sending the create request. 

If we feel strongly about this, we'd have to add an endpoint to the enterprise API in order to retrieve an image by name so we can use the image-id in the create environment request.

The clilent will require the image-id either way.

Passing a name would be nice but I don't think taking an image id is out of the ordinary(Docker commands take container id's).

We could also have a `coder images ls` in the future that helps make obtaining these image id's more convenient.

cmoog: Yeah I was thinking we'd have something like `findEnv` for images. We've been pretty good so far about using human friendly values so let's stick to that before unhiding this. Makes sense to tackle that in a follow-up PR though.

ci/steps/gendocs.sh
```diff
@@ -13,7 +13,7 @@ go run ./cmd/coder gen-docs ./docs
 
 # remove cobra footer from each file
 for filename in ./docs/*.md; do
-  trimmed=$(head -n -1 "$filename")
+  trimmed=$(head -n $(( $(wc -l "$filename" | awk '{print $1}') - 1 )) "$filename")
   echo "$trimmed" >$filename
 done
```
cmoog, line 18: We can delete this whole block now 😄

internal/cmd/envs.go
```diff
@@ -125,3 +134,78 @@ coder envs --user charlie@coder.com ls -o json \
...
+	)
+
+	cmd := &cobra.Command{
+		Use:    "create [environment_name]",
+		Short:  "create a new environment.",
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Create a new environment under the active user.",
+		Example: `# create a new environment using default resource amounts
+coder envs create --image id-of-imported-image my-env-name
```
cmoog, line 153: The example should be "runnable" in theory for some data. Let's use a fake image id here.

internal/cmd/envs.go
```diff
@@ -125,3 +134,78 @@ coder envs --user charlie@coder.com ls -o json \
...
+
+# create a new environment using custom resource amounts
+coder envs create --cores 4 --disk 100 --memory 8 --image id-of-imported-image --org id-of-existing-organization my-env-name
```
cmoog, line 156: should be `--cpu`

internal/cmd/envs.go
```diff
@@ -125,3 +134,78 @@ coder envs --user charlie@coder.com ls -o json \
...
+
+# using short-hand flags.
+coder envs create -c 4 -d 100 -m 8 -i id-of-imported-image -o id-of-existing-organization my-env-name`,
```
cmoog, line 159: Don't think we need an example for short flags.

cmoog (approved):
Just a few fix-ups then it's good to go.

PR #150, opened 2020-10-21
Add new rich error messages

cmoog:
Inspired by the `cargo` message formatting.

Screen Shot 2020-10-21 at 1 08 31 PM

Screen Shot 2020-10-21 at 1 34 53 AM
Screen Shot 2020-10-21 at 1 13 48 PM

### DIFF
--- a/internal/clog/error.go
+++ b/internal/clog/error.go
@@ -1,0 +1,131 @@
+package clog
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/fatih/color"
+	"golang.org/x/xerrors"
+)
+
+// CLIMessage provides a human-readable message for CLI errors and messages.
+type CLIMessage struct {
+	Level  string
+	Color  color.Attribute
+	Header string
+	Lines  []string
+}
+
+// CLIError wraps a CLIMessage and allows consumers to treat it as a normal error.
+type CLIError struct {
+	CLIMessage
+	error
+}
+
+// String formats the CLI message for consumption by a human.
+func (m CLIMessage) String() string {
+	var str strings.Builder
+	str.WriteString(fmt.Sprintf("%s: %s\n",
+		color.New(m.Color).Sprint(m.Level),
+		color.New(color.Bold).Sprint(m.Header)),
+	)
+	for _, line := range m.Lines {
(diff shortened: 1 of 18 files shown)


internal/cmd/shell.go
```diff
@@ -188,13 +189,26 @@ func runCommand(ctx context.Context, envName, command string, args []string) err
...
 	return nil
 }
 
+func networkErr(client *coder.Client, env *coder.Environment) error {
+	if env.LatestStat.ContainerStatus != coder.EnvironmentOn {
+		return clog.Fatal(
+			"environment is not running",
+			fmt.Sprintf("environment %q is not running", env.Name),
+			fmt.Sprintf("its current status is %q", env.LatestStat.ContainerStatus),
+			"",
```
coadler, line 205: The empty string on a lot of these are a bit confusing, I know it's a bit redundant but maybe something like `clog.LineSeparator` or similar would make it clear why it's used.

internal/clog/error.go
```diff
@@ -0,0 +1,133 @@
...
+func Log(err error) {
+	var cliErr CLIError
+	if !xerrors.As(err, &cliErr) {
+		cliErr = Fatal(err.Error())
+	}
+	fmt.Fprintln(os.Stderr, cliErr.String())
+}
+
+// LogInfo prints the given info message to stderr.
+func LogInfo(header string, lines ...string) {
```
coadler, line 52: I'm a bit confused why `LogInfo` and `LogSuccess` are the only ones prefixed with `Log`

coadler (commented):
Looks really good!

internal/clog/error.go
cmoog: Yeah this is definitely confusing. The issue I'm having is that `Error`, `Warn`, and `Fatal` return errors. So it would potentially even more confusing for these to be the only ones with side-effects. 

It gets pretty tricky when we actually _do_ want to log errors as they happen: for example when iterating through a series of `stop` actions.

Any ideas?

internal/cmd/shell.go
cmoog: Oh nice nice.

internal/clog/error.go
coadler: Ohh i see now, I was skimming over the returns. Considering the others return errors this makes more sense

coadler (approved):

PR #145, opened 2020-10-19
Add coder envs rm command for removing environments

cmoog:
Adds a new command
```
coder envs rm [...environment_names] [flags]
```

PR #151, opened 2020-10-21
Add coder envs edit command

fuskovic:
# Usage

Adds the `edit` command to the `envs` command.

```         
Edit an existing environment owned by the active user.

Usage:
  coder envs edit [flags]

Examples:
coder envs edit back-end-env --cores 4

coder envs edit back-end-env --disk 20

Flags:
  -c, --cpu float32      The number of cpu cores the environment should be provisioned with.
  -d, --disk int         The amount of disk storage an environment should be provisioned with.
      --follow           follow buildlog after initiating rebuild
  -g, --gpu int          The amount of disk storage to provision the environment with.
  -h, --help             help for edit
  -i, --image string     image ID of the image you wan't the environment to be based off of.
  -m, --memory float32   The amount of RAM an environment should be provisioned with.
  -t, --tag string       image tag of the image you wan't to base the environment off of. (default "latest")
```
# Example

Screen Shot 2020-10-22 at 7 24 33 AM

### DIFF
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -72,6 +72,7 @@ func envsCommand() *cobra.Command {
 	}
 	lsCmd.Flags().StringVarP(&outputFmt, "output", "o", "human", "human | json")
 	cmd.AddCommand(lsCmd)
+	cmd.AddCommand(editEnvCommand(&user))
 	cmd.AddCommand(stopEnvCommand(&user))
 	cmd.AddCommand(watchBuildLogCommand())
 	cmd.AddCommand(rebuildEnvCommand())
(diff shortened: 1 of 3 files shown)


internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Edit an existing environment owned by the active user.",
+		Example: `coder envs edit back-end-env
+# increase number of cores for back-end env
+coder envs edit back-end-env --cores 4
+
+# increase disk storage for back-end-env
+coder envs edit back-end-env --disk 20
+`,
```
cmoog, line 88: don't need the trailing newline

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+	)
+
+	editEnvCmd := &cobra.Command{
+		Use:    "edit",
+		Short:  "edit an existing environment owned by the active user.",
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Edit an existing environment owned by the active user.",
+		Example: `coder envs edit back-end-env
+# increase number of cores for back-end env
```
cmoog, line 83: not necessarily increasing. this one seems pretty self-explanatory such that we can remove the comment.

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+coder envs edit back-end-env --cores 4
+
+# increase disk storage for back-end-env
```
cmoog, line 86: same here

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+		img      string
+		tag      string
+		cpuCores float32
+		memGB    float32
+		diskGB   int
+		gpus     int
+		services []string
+	)
+
+	editEnvCmd := &cobra.Command{
```
cmoog, line 76: Can we move this into its own function? The `lsCmd` is total garbage and is getting fixed as part of https://github.com/cdr/coder-cli/pull/145

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+
+# increase disk storage for back-end-env
+coder envs edit back-end-env --disk 20
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := newClient()
+			if err != nil {
+				return err
+			}
+			envs, err := getEnvs(cmd.Context(), client, user)
```
cmoog, line 94: use `findEnv` instead

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+						updateReq.ImageTag = &tag
+					}
+
+					if err := client.EditEnvironment(cmd.Context(), e.ID, updateReq); err != nil {
+						flog.Error("failed to apply changes to environment: '%s'", name)
+						return xerrors.Errorf("failed to apply changes to environment: '%s'", name)
+					}
+					flog.Success("successfully applied changes to environment: '%s'", name)
+					return nil
+				}
```
cmoog, line 160: same issue here as in https://github.com/cdr/coder-cli/pull/149. More accurate message and the `--follow` flag will help reduce confusion. Now that we have the `clog` PR merged you can even add something like
```go
clog.LogSuccess(
    "applied changes to the environment, rebuilding...", 
    clog.Tip(`run "coder envs watch-build %q" to trail the build logs`),
),  
```

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+					}
+
+					if tag == "" {
+						updateReq.ImageTag = &e.ImageTag
+					} else {
+						updateReq.ImageTag = &tag
+					}
+
+					if err := client.EditEnvironment(cmd.Context(), e.ID, updateReq); err != nil {
+						flog.Error("failed to apply changes to environment: '%s'", name)
```
cmoog, line 155: No need to ever log an error like this _unless_ we're iterating and performing fail-able operations that we want handled at the call site. Otherwise, the error returned by this func will be propagated and logged.

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+			return xerrors.Errorf("environment '%s' not found", name)
+		},
+	}
+	editEnvCmd.Flags().StringVarP(&img, "image", "i", "", "image ID of the image you wan't the environment to be based off of.")
+	editEnvCmd.Flags().StringVarP(&tag, "tag", "t", "latest", "image tag of the image you wan't to base the environment off of.")
+	editEnvCmd.Flags().Float32P("cores", "c", cpuCores, "The number of cpu cores the environment should be provisioned with. (optional: true, supports fractional amounts)")
+	editEnvCmd.Flags().Float32P("memory", "m", memGB, "The amount of RAM an environment should be provisioned with. (optional: true, supports fractional amounts)")
+	editEnvCmd.Flags().IntP("disk", "d", diskGB, "The amount of disk storage an environment should be provisioned with. (optional: true, supports whole numbers only)")
+	editEnvCmd.Flags().IntP("gpu", "g", gpus, "The amount of disk storage to provision the environment with. (optional: true, supports whole numbers only)")
+	editEnvCmd.Flags().StringSliceP("services", "s", services, "The services that the environment should be initialized with. (optional: true)")
```
cmoog, line 171: same here

internal/cmd/envs.go
```diff
@@ -62,8 +62,115 @@ func envsCommand() *cobra.Command {
...
+					flog.Success("successfully applied changes to environment: '%s'", name)
+					return nil
+				}
+			}
+			return xerrors.Errorf("environment '%s' not found", name)
+		},
+	}
+	editEnvCmd.Flags().StringVarP(&img, "image", "i", "", "image ID of the image you wan't the environment to be based off of.")
+	editEnvCmd.Flags().StringVarP(&tag, "tag", "t", "latest", "image tag of the image you wan't to base the environment off of.")
+	editEnvCmd.Flags().Float32P("cores", "c", cpuCores, "The number of cpu cores the environment should be provisioned with. (optional: true, supports fractional amounts)")
```
cmoog, line 167: `--cpu`

cmoog (changes requested):

internal/cmd/envs.go
cmoog: Also, we should probably error if they call `coder envs edit [env]` and suggest that they use `coder envs rebuild [env]` instead.

ci/steps/gendocs.sh
```diff
@@ -13,7 +13,7 @@ go run ./cmd/coder gen-docs ./docs
 
 # remove cobra footer from each file
 for filename in ./docs/*.md; do
-  trimmed=$(head -n -1 "$filename")
+  trimmed=$(head -n $(( $(wc -l "$filename" | awk '{print $1}') - 1 )) "$filename")
   echo "$trimmed" >$filename
 done
```
cmoog, line 18: Can remote this now.

internal/cmd/envs.go
```diff
@@ -125,3 +126,111 @@ coder envs --user charlie@coder.com ls -o json \
...
+		follow   bool
+	)
+
+	cmd := &cobra.Command{
+		Use:    "edit",
+		Short:  "edit an existing environment owned by the active user.",
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		Long:   "Edit an existing environment owned by the active user.",
+		Example: `coder envs edit back-end-env --cores 4
```
cmoog, line 147: `--cpu`

cmoog (approved):

PR #153, opened 2020-10-22
Fix ioctl issue when using stdin with coder sh

cmoog:


jawnsy:
@cmoog this looks useful and low risk, any reason we shouldn't merge this soon?

cmoog:
This isn't an issue anymore now that we shell out to `ssh` directly.

PR #155, opened 2020-10-23
Trail build logs if follow enabled for envs edit

fuskovic:
### What this does

Trails build logs when `--follow` is enabled for `coder envs edit`


internal/cmd/envs.go
```diff
@@ -298,6 +298,12 @@ coder envs edit back-end-env --disk 20`,
 				clog.BlankLine,
 				clog.Tipf(`run "coder envs watch-build %q" to trail the build logs`, envName),
```
cmoog, line 299: Shouldn't we only print this if `!follow`?

docs/coder.md
```diff
@@ -2,10 +2,6 @@
 
 coder provides a CLI for working with an existing Coder Enterprise installation
 
-### Synopsis
-
-coder provides a CLI for working with an existing Coder Enterprise installation
-
```
coadler, line 8: Can we have a separate commit explaining why these were removed? Odd for it to be bundled into the other commit.

coadler (approved):

docs/coder.md
cmoog: Looks like it's because we updated `cobra`. CI enforces that the docs are properly generated and always up-to-date.

coadler: Ah, didn't realize this was generated. sg

cmoog (approved):

PR #158, opened 2020-10-24
Show warning on coder-cli / Coder API version mismatch

cmoog:
Show a warning to the user if the `<major>.<minor>` versions do not match between the CLI and Coder API. 

Screen Shot 2020-10-24 at 6 30 10 PM

## Testing

Trying running a build generated from `./ci/steps/build.sh` and see that the version is injected and the warning hidden.


internal/cmd/auth.go
```diff
@@ -52,5 +53,14 @@ func newClient() (*coder.Client, error) {
 		return nil, err
 	}
 
+	if !version.VersionsMatch(apiVersion) {
+		clog.Log(clog.Warn(
+			"version mismatch detected",
+			fmt.Sprintf("coder-cli version: %s", version.Version),
+			fmt.Sprintf("Coder API version: %s", apiVersion), clog.BlankLine,
+			clog.Tipf("download the appropriate version here: https://github.com/cdr/coder-cli/releases"),
```
coadler, line 61: Just to make sure, this is thrown to stderr correct?

coadler (approved):
New logging makes this look amazing

internal/cmd/auth.go
cmoog: Yup, all log levels are thrown in stderr.

PR #159, opened 2020-10-26
Change coder create | edit envs image flag to take image name and source defaults from image

fuskovic:
# What this does

- Users can pass the image name to the `--image` flag instead of having to pass the image id.
- This applies to both coder `create` and `edit`
- Coder `create` and `edit` both now source default resource amounts based on the imported image vs arbitrary ones we had previously defined in the CLI.

# Create

## Create environment with default resource amounts

This screenshot also demonstrates some examples of errors users might have.

create

## Create environment with custom resource amounts

create_w_custom

# Edit

## Change the image of an environment

When an environment image is changed and resource amounts are left unspecified, the resource amounts will be changed to that of the default resource amounts specified by the new image the environment will be using.

Also not there's a caveat here in the screenshot it's worth being aware of. Notice when we are going from `codercom/enterprise-dev` to `ubuntu` that we get a warning message about how we can't reduce the amount of disk storage. The ubuntu default is 10 so the tool is smart enough to recognize that the default amount of the new image for disk is smaller than what it's currently using and keeps it the same while warning the user.

edit_image

## Change the resource amounts of an environment

edit_resoures

## Change both the image and resource amounts of an environment

edit_both

### DIFF
--- a/coder-sdk/image.go
+++ b/coder-sdk/image.go
@@ -13,7 +13,7 @@ type Image struct {
 	Description     string  `json:"description"`
 	URL             string  `json:"url"` // User-supplied URL for image.
 	DefaultCPUCores float32 `json:"default_cpu_cores"`
-	DefaultMemoryGB int     `json:"default_memory_gb"`
+	DefaultMemoryGB float32 `json:"default_memory_gb"`
 	DefaultDiskGB   int     `json:"default_disk_gb"`
 	Deprecated      bool    `json:"deprecated"`
 }
@@ -46,4 +46,13 @@ func (c Client) ImportImage(ctx context.Context, orgID string, req ImportImageRe
 		return nil, err
 	}
 	return &img, nil
+}
+
+// OrganizationImages returns all of the images imported for orgID.
+func (c Client) OrganizationImages(ctx context.Context, orgID string) ([]Image, error) {
+	var imgs []Image
+	if err := c.requestBody(ctx, http.MethodGet, "/api/orgs/"+orgID+"/images", nil, &imgs); err != nil {
+		return nil, err
+	}
+	return imgs, nil
 }
--- a/internal/clog/error.go
+++ b/internal/clog/error.go
@@ -68,6 +68,16 @@ func LogSuccess(header string, lines ...string) {
 	}.String())
 }
 
+// LogWarn prints the given warn message to stderr.
+func LogWarn(header string, lines ...string) {
+	fmt.Fprint(os.Stderr, CLIMessage{
+		Level:  "warning",
+		Color:  color.FgYellow,
+		Header: header,
+		Lines:  lines,
+	}.String())
+}
+
 // Warn creates an error with the level "warning".
 func Warn(header string, lines ...string) CLIError {
 	return CLIError{
--- a/internal/cmd/ceapi.go
+++ b/internal/cmd/ceapi.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"cdr.dev/coder-cli/coder-sdk"
 	"cdr.dev/coder-cli/internal/clog"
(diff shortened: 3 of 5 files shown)


internal/cmd/ceapi.go
```diff
@@ -81,3 +82,94 @@ func findEnv(ctx context.Context, client *coder.Client, envName, userEmail strin
...
+	}
+
+	if numImgs > 1 {
+		var lines []string
+
+		for _, img := range userImgs {
+			lines = append(lines, img.Repository)
+		}
+
+		clog.LogInfo(
```
cmoog, line 105: This needs to be returned as an error, use `clog.Fatal`.

internal/cmd/ceapi.go
```diff
@@ -81,3 +82,94 @@ func findEnv(ctx context.Context, client *coder.Client, envName, userEmail strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	orgs = lookupUserOrgs(u, orgs)
+	var importedImgs []*coder.Image
+
+	// Get all of the imported images for all of the orgs the user belongs to.
+	for _, org := range orgs {
+		imgs, err := client.GetOrganizationImages(ctx, org.ID)
```
cmoog, line 131: This is a bit of a confusing API. Am I correct in saying that the organization in which the environment is created is essentially inferred from which org has the image they specified? If the user is a member of multiple orgs, the org should be a required flag.

internal/cmd/ceapi.go
```diff
@@ -81,3 +82,94 @@ func findEnv(ctx context.Context, client *coder.Client, envName, userEmail strin
...
+		if err != nil {
+			return nil, err
+		}
+		importedImgs = append(importedImgs, imgs...)
+	}
+	return importedImgs, nil
+}
+
+// returns all images that contain imgName as a substring and have been imported by the user.
+func lookupUserImgs(ctx context.Context, client *coder.Client, email, imgName string) ([]*coder.Image, error) {
```
cmoog, line 141: This is quite confusing behavior for this function. It should either return a single image if there is an exact match or an err. To include the "help" information in the error use `clog.Fatal`.

cmoog (changes requested):

fuskovic:
@tychoish 

It looks like all the envs commands take pointers because the value of `user` is propogated by the persistent flag `user` upstream in the parent command. Other than that I pushed up all your suggested changes.

@cmoog 

Pushed all yours up as well and I updated the PR description to reflect any new changes and behaviors.

internal/cmd/ceapi.go
```diff
@@ -81,3 +82,111 @@ func findEnv(ctx context.Context, client *coder.Client, envName, userEmail strin
 		clog.Tipf("run \"coder envs ls\" to view your environments"),
 	)
 }
+
+func findImg(ctx context.Context, client *coder.Client, email, imgName, orgName string) (*coder.Image, error) {
```
tychoish, line 86: same thing with args struct here, just for clarity.

tychoish (approved):

internal/clog/error.go
```diff
@@ -68,6 +68,16 @@ func LogSuccess(header string, lines ...string) {
 	}.String())
 }
 
+// LogWarn prints the given warn message to stderr.
+func LogWarn(header string, lines ...string) {
```
cmoog, line 72: Nice 👍

coder-sdk/image.go
```diff
@@ -47,3 +47,12 @@ func (c Client) ImportImage(ctx context.Context, orgID string, req ImportImageRe
 	}
 	return &img, nil
 }
+
+// GetOrganizationImages returns all of the images imported for orgID.
+func (c Client) GetOrganizationImages(ctx context.Context, orgID string) ([]Image, error) {
```
cmoog, line 52: can drop the `Get` here to match the other funcs.

internal/cmd/ceapi.go
```diff
@@ -81,3 +82,118 @@ func findEnv(ctx context.Context, client *coder.Client, envName, userEmail strin
...
+
+		if conf.orgName == "" {
+			// if orgName is unset we know the user is only part of one org.
+			return imgs, nil
+		}
+	}
+	return nil, xerrors.Errorf("org name %q not found", conf.orgName)
+}
+
+func isMultiOrgMember(ctx context.Context, email string) (*bool, error) {
```
cmoog, line 181: why not just `(bool, error)`?

internal/cmd/envs.go
```diff
@@ -387,3 +404,110 @@ func rmEnvsCommand(user *string) *cobra.Command {
...
+	if conf.diskGB == 0 {
+		updateReq.DiskGB = &defaultDiskGB
+	} else {
+		updateReq.DiskGB = &conf.diskGB
+	}
+
+	// Environment disks can not be shrink so we have to overwrite this
+	// if the user accidentally requests it or if the default diskGB value for a
+	// newly requested image is smaller than the current amount the environment is using.
+	if *updateReq.DiskGB < conf.environment.DiskGB {
```
cmoog, line 492: Nice 👍

cmoog (approved):

internal/cmd/ceapi.go
fuskovic: `false` would imply that the user that `email` belongs to is not a multi-org member and that may not be true if we fail to create a new client or get the user by email.

cmoog: Yeah that's fair. I guess it really comes down to a failure of the language but it is certainly a `go` convention to ignore the first value if `err != nil`.

cmoog:
Also- I've been moving towards `Squash and merge`. Makes the commit history on master super clean.

internal/cmd/envs.go
```diff
@@ -169,27 +164,61 @@ coder envs create --image 5f443b16-30652892427b955601330fa5 my-env-name
...
+			}
+
+			if multiOrgMember && org == "" {
+				return xerrors.New("org is required for multi-org members")
+			}
+
+			client, err := newClient()
+			if err != nil {
+				return err
+			}
```
cmoog, line 182: Just noticed this. `isMultiOrgMember` should accept `client` as an argument. `newClient` will log a warning if the versions mismatch so we wouldn't want that to happen twice. It also makes a network call to see if the auth is valid which is wasteful to have twice.

PR #163, opened 2020-10-29
Error coder sh commands if the env requires a rebuild

cmoog:
https://app.clubhouse.io/coder/story/352/reject-or-warn-on-coder-sh-when-an-environment-needs-to-be-rebuilt

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -21,24 +21,24 @@ func Make() *cobra.Command {
 	}
 
 	app.AddCommand(
-		makeLoginCmd(),
-		makeLogoutCmd(),
+		loginCmd(),
+		logoutCmd(),
 		shCmd(),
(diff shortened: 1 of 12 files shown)


internal/cmd/cmd.go
```diff
@@ -23,7 +23,7 @@ func Make() *cobra.Command {
 	app.AddCommand(
 		makeLoginCmd(),
 		makeLogoutCmd(),
-		makeShellCmd(),
+		shCmd(),
```
deansheather, line 26: Why the rename of this particular command? Consistency seems better here, so we should change them all or not change this one.

deansheather (approved):

internal/cmd/cmd.go
cmoog: Yeah fair point. Just updated all uses of `make*`

PR #171, opened 2020-11-03
Hide secrets command

fuskovic:


cmoog:
Let's also print a warning message saying that it's deprecated.

ci/integration/secrets_test.go
```diff
@@ -1,81 +1,80 @@
...
-	"cdr.dev/coder-cli/pkg/tcli"
+	//"cdr.dev/coder-cli/pkg/tcli"
 )
 
 func TestSecrets(t *testing.T) {
-	t.Parallel()
-	run(t, "secrets-cli-tests", func(t *testing.T, ctx context.Context, c *tcli.ContainerRunner) {
-		headlessLogin(ctx, t, c)
+	// t.Parallel()
+	// run(t, "secrets-cli-tests", func(t *testing.T, ctx context.Context, c *tcli.ContainerRunner) {
```
cmoog, line 13: let's call `t.Skip` rather than commenting them out

internal/cmd/secrets.go
```diff
@@ -17,9 +17,19 @@ import (
...
+		Long:   "Interact with secrets objects owned by the active user.",
+		Hidden: true,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			clog.LogWarn(
+				"The 'secrets' command is now deprecated",
+				"It will be removed in the next minor release",
+			)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			_ = cmd.Help()
```
cmoog, line 31: is this needed?

cmoog (approved):

internal/cmd/secrets.go
fuskovic: I noticed the `persistentPreRun` wouldn't run for `coder secrets` without a `Run`.

Once I added it then it worked but only printed the warning in the `persistentPreRun` so I added this particular line to make sure the help was there too.

PR #173, opened 2020-11-03
Add notarization of MacOS binary to CI

### LINKED ISSUE #26: macOS / Darwin release not signed
The binary release for the Darwin CLI isn't signed, and causes Gatekeeper to throw warnings when run. This is an issue in our environment as we require all installed binaries to be signed for security.

kylecarbs:
Closes #26


cmoog (approved):
💯

PR #176, opened 2020-11-04
Enable more linting rules

cmoog:


cmd/coder/main.go
```diff
@@ -31,16 +30,19 @@ func main() {
...
 	app := cmd.Make()
 	app.Version = fmt.Sprintf("%s %s %s/%s", version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 
 	if err := app.ExecuteContext(ctx); err != nil {
 		clog.Log(err)
+		restoreTerminal()
 		os.Exit(1)
 	}
+	restoreTerminal()
+	cancel()
```
cmoog, line 47: I was very skeptical of `gocritic` but the fact that it caught this bug was enough to convince me.

kylecarbs (approved):
Kinda impressed

PR #182, opened 2020-11-08
Release action

cmoog:
This PR adds a new Github Action workflow that runs whenever `v*` tags are pushed. It'll create a release build and upload the assets to a drafted Github release. This will still require that a human intervenes to make the release public and add any pertinent release notes.


deansheather (approved):
LGTM, needs a change to our enterprise release script so we avoid making two releases.

coadler (approved):
Obv didn't test but LGTM if it works. I assume the draft release just shows up to repo admins and we can go in and publish?

cmoog:
> I assume the draft release just shows up to repo admins and we can go in and publish?

Yup, exactly.

PR #191, opened 2020-11-16
Use dedicated Column interface for table formatting nested structs

cmoog:
For some background: we've been implementing `Stringer` to specify formatting for table columns in cases where the row has a nested struct. I realized that his may cause unexpected behavior for public consumers of `coder-sdk` when logging or otherwise formatting API types. 

This PR introduces a custom `Column` interface for specifying the format of a nested struct that should be represented as a single column.

### DIFF
--- a/pkg/tablewriter/tablewriter.go
+++ b/pkg/tablewriter/tablewriter.go
@@ -10,6 +10,11 @@ import (
 
 const structFieldTagKey = "table"
 
+// Column defines an interface for formatting a type as a column.
+type Column interface {
+	Column() string
+}
+
 // StructValues tab delimits the values of a given struct.
 //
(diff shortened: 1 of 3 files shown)


pkg/tablewriter/tablewriter.go
```diff
@@ -20,7 +25,12 @@ func StructValues(data interface{}) string {
 		if shouldHideField(v.Type().Field(i)) {
 			continue
 		}
-		fmt.Fprintf(s, "%v\t", v.Field(i).Interface())
+		value := v.Field(i).Interface()
+		if column, ok := value.(Column); ok {
```
cmoog, line 29: In principle, I hate this pattern.

cmoog:
Might revisit this at some point.

PR #195, opened 2020-12-02
cli: Update UXState from config-ssh

vapurrmaid:
## Context

In optimizing the dashboard environment experiences, tracking whether a user has run config-ssh locally enables personalized education about the feature.

## Testing

1. I downloaded and ran `coder config-ssh` using the binary attached to the CI run
2. I verified that my ux state was updated

### DIFF
--- a/coder-sdk/users.go
+++ b/coder-sdk/users.go
@@ -114,6 +114,14 @@ func (c Client) UpdateUser(ctx context.Context, userID string, req UpdateUserReq
 	return c.requestBody(ctx, http.MethodPatch, "/api/private/users/"+userID, req, nil)
 }
 
+// UpdateUXState applies a partial update of the user's UX State.
+func (c Client) UpdateUXState(ctx context.Context, userID string, uxsPartial map[string]interface{}) error {
+	if err := c.requestBody(ctx, http.MethodPut, "/api/private/users/"+userID+"/ux-state", uxsPartial, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
 // CreateUserReq defines the request parameters for creating a new user resource.
 type CreateUserReq struct {
 	Name              string    `json:"name"`
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"cdr.dev/coder-cli/pkg/clog"
+
 	"cdr.dev/coder-cli/coder-sdk"
 	"cdr.dev/coder-cli/internal/config"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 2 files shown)


coder-sdk/users.go
```diff
@@ -46,6 +46,14 @@ func (c Client) Me(ctx context.Context) (*User, error) {
 	return c.UserByID(ctx, Me)
 }
 
+// MyUXState updates the authenticated user's UX State.
+func (c Client) MyUXState(ctx context.Context, uxsPartial map[string]interface{}) error {
```
cmoog, line 50: Maybe this would be better as `UpdateUXState` where it accepts a `userID string` param. Then we can pass `coder.Me` where it's used.

internal/cmd/configssh.go
```diff
@@ -135,6 +135,7 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 			fmt.Printf("Your private ssh key was written to \"%s\"\n", privateKeyFilepath)
 		}
 
+		writeSSHUXState(ctx, client)
```
cmoog, line 138:
```suggestion
		_ = writeSSHUXState(ctx, client)
```

internal/cmd/configssh.go
```diff
@@ -215,3 +216,12 @@ func readStr(filename string) (string, error) {
 	}
 	return string(contents), nil
 }
+
+func writeSSHUXState(ctx context.Context, client *coder.Client) {
+	// Update UXState that coder config-ssh has been run by the currently
+	// authenticated user
+	err := client.MyUXState(ctx, map[string]interface{}{"cliSSHConfigured": true})
+	if err != nil {
+		fmt.Println("⚠ The Coder web client may not recognize that you've configured SSH.")
```
cmoog, line 225: `clog.LogWarn` could provide some nice formatting for this.

cmoog (changes requested):

internal/cmd/configssh.go
vapurrmaid: This caused a CI error - `writeSSHUXState` as written doesn't return a value.

I initially had it return an err, but go complexity complained about too many branches.

cmoog: Oooh my bad, misread that.

vapurrmaid: It's all good - was totally my initial intention too

internal/cmd/configssh.go
vapurrmaid: Awesome stuff :+1:

coder-sdk/users.go
```diff
@@ -114,6 +114,14 @@ func (c Client) UpdateUser(ctx context.Context, userID string, req UpdateUserReq
 	return c.requestBody(ctx, http.MethodPatch, "/api/private/users/"+userID, req, nil)
 }
 
+// UpdateUXState updates the authenticated user's UX State.
```
cmoog, line 117:
```suggestion
// UpdateUXState applies a partial update of the user's UX State.
```

cmoog: Only getting particular because these are public on `godoc`.

cmoog (approved):
Nice 👍

coder-sdk/users.go
vapurrmaid: Hey mr @cmoog , I appreciate fine :eyes: - thanks so much :pray:

PR #197, opened 2020-12-03
Use env.ssh_available field in config-ssh

deansheather:
Instead of dialing the SSH port to determine if SSH is available, uses a field exposed from cemanager in environments responses. This field has been added for the UI, but it makes sense to use it here too.

https://github.com/cdr/c/pull/6108


internal/cmd/configssh.go
```diff
@@ -145,6 +144,17 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 	}
 }
 
+// sshAvailable returns true if SSH is available for at least some environments.
```
vapurrmaid, line 147:
```suggestion
// sshAvailable returns true if SSH is available for at least one environment.
```

internal/cmd/configssh.go
```diff
@@ -181,20 +195,6 @@ func makeSSHConfig(host, userName, envName, privateKeyFilepath string) string {
 `, envName, host, userName, envName, privateKeyFilepath)
 }
 
-func isSSHAvailable(ctx context.Context) bool {
-	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
-	defer cancel()
-
-	host, err := configuredHostname()
```
vapurrmaid, line 188: I can't recall, can the `configuredHostname` func be removed as a result of this? May have been the only spot it was called

internal/cmd/configssh.go
```diff
@@ -109,6 +103,11 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
 		if len(envs) < 1 {
 			return xerrors.New("no environments found")
 		}
+
+		if !sshAvailable(envs) {
+			return xerrors.New("SSH is disabled or not available any environments in your Coder Enterprise deployment.")
```
cmoog, line 108:
```suggestion
			return xerrors.New("SSH is disabled or not available for any environments in your Coder Enterprise deployment.")
```

cmoog (approved):
LGTM

cmoog (approved):
Appreciate the cleanup 👍 . Looking at the failing integration test, I'd say the best course of action is to merge this as-is once https://github.com/cdr/c/pull/6108 is ready, then update the submodule in the `cdr/c` PR.

We can make sure the integration tests pass once master is deployed.

PR #201, opened 2020-12-05
Add cvm flag to CLI and fields to coder-sdk

cmoog:
[ch4260]

### DIFF
--- a/docs/coder_envs_edit.md
+++ b/docs/coder_envs_edit.md
@@ -21,15 +21,17 @@ coder envs edit back-end-env --disk 20
 ### Options
 
 ```
-  -c, --cpu float32      The number of cpu cores the environment should be provisioned with.
-  -d, --disk int         The amount of disk storage an environment should be provisioned with.
-      --follow           follow buildlog after initiating rebuild
-  -g, --gpu int          The amount of disk storage to provision the environment with.
-  -h, --help             help for edit
-  -i, --image string     name of the image you want the environment to be based off of.
-  -m, --memory float32   The amount of RAM an environment should be provisioned with.
-  -o, --org string       name of the organization the environment should be created under.
-  -t, --tag string       image tag of the image you want to base the environment off of. (default "latest")
+      --container-vm       deploy the environment as a Container-based VM
+  -c, --cpu float32        The number of cpu cores the environment should be provisioned with.
+  -d, --disk int           The amount of disk storage an environment should be provisioned with.
+      --follow             follow buildlog after initiating rebuild
+  -g, --gpu int            The amount of disk storage to provision the environment with.
+  -h, --help               help for edit
+  -i, --image string       name of the image you want the environment to be based off of.
+  -m, --memory float32     The amount of RAM an environment should be provisioned with.
+      --not-container-vm   do not deploy the environment as a Container-based VM
+  -o, --org string         name of the organization the environment should be created under.
+  -t, --tag string         image tag of the image you want to base the environment off of. (default "latest")
 ```
 
 ### Options inherited from parent commands
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -148,6 +148,7 @@ func createEnvCmd(user *string) *cobra.Command {
 		img    string
 		tag    string
 		follow bool
+		useCVM bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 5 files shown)


cmoog:
`--not-container-vm` is pretty bad so please note if you have a better alternative.

kylecarbs:
Thoughts on just `--container`?

internal/cmd/envs.go
```diff
@@ -401,6 +412,16 @@ func buildUpdateReq(ctx context.Context, client *coder.Client, conf updateConf)
...
+	if conf.useCVM && conf.noCVM {
+		return nil, xerrors.New("--container-vm and --not-container-vm flags conflict")
+	}
+	if conf.useCVM {
+		updateReq.UseContainerVM = boolP(true)
+	}
+	if conf.noCVM {
+		updateReq.UseContainerVM = boolP(false)
+	}
+
```
fuskovic, line 424: Scenario:

I have an existing CVM enabled env.

I wan't to update memory.

I don't wan't to disable CVM feature on said env but if I don't pass `--container-vm` again the update will assume that I do?

Does this mean I have to pass `--container-vm` every time I wan't to make an environment update but not toggle using the CVM itself?

If so, I believe this burden should fall on the program and not the user. 

It's not fair to force the user to be explicit or have to remember special details like this.

My understanding is that when we attempt to take an env from cvm-enabled -> cvm-disabled and vice versa the environment fails to rebuild successfully anyway ( correct me if i'm wrong )

If that's correct, should we be allowed to toggle this on env updates at all?

fuskovic (changes requested):

docs/coder_envs_edit.md
```diff
@@ -21,15 +21,17 @@ coder envs edit back-end-env --disk 20
...
-  -t, --tag string       image tag of the image you want to base the environment off of. (default "latest")
+      --container-vm       deploy the environment as a Container-based VM
+  -c, --cpu float32        The number of cpu cores the environment should be provisioned with.
+  -d, --disk int           The amount of disk storage an environment should be provisioned with.
+      --follow             follow buildlog after initiating rebuild
+  -g, --gpu int            The amount of disk storage to provision the environment with.
+  -h, --help               help for edit
+  -i, --image string       name of the image you want the environment to be based off of.
+  -m, --memory float32     The amount of RAM an environment should be provisioned with.
+      --not-container-vm   do not deploy the environment as a Container-based VM
```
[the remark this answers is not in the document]
cmoog: Because this is the edit command where it's a partial update. Right?

internal/cmd/envs.go
```diff
@@ -328,6 +333,8 @@ coder envs edit back-end-env --disk 20`,
 	cmd.Flags().IntVarP(&disk, "disk", "d", 0, "The amount of disk storage an environment should be provisioned with.")
 	cmd.Flags().IntVarP(&gpus, "gpu", "g", 0, "The amount of disk storage to provision the environment with.")
 	cmd.Flags().BoolVar(&follow, "follow", false, "follow buildlog after initiating rebuild")
+	cmd.Flags().BoolVar(&useCVM, "container-vm", false, "deploy the environment as a Container-based VM")
+	cmd.Flags().BoolVar(&notCVM, "not-container-vm", false, "do not deploy the environment as a Container-based VM")
```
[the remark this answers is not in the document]
cmoog: Same as above, this is the edit command so the update is partial.

internal/cmd/envs.go
cmoog: I think the previous comments should resolve these concerns?

fuskovic: I understand the partial update. It's not obvious to me why we are allowing the user to toggle enabling cvm on an env if that will fail the rebuild?

cmoog: Ah I see, this will not be the behavior shortly. We're going to have upgrading/downgrading working properly shortly.

fuskovic: Ah ok well that's good.

What do you think about a single flag for toggling state where if you use it the value of `UseCVM` becomes `!env.UseCVM` and maybe prompt the user about the implications of switching or even prompt for confirmation? Trying to think of alternatives

cmoog: Toggle flags are not commonly used for cases like this as it can cause super odd behavior based on prior state. Running the same edit command twice though yield the same result.

Having a positive and negative flag, although not ideal, feels like the correct approach here.

cmoog: > If so, I believe this burden should fall on the program and not the user.

The partial update behavior takes care of this...

cmoog:
> Thoughts on just --container?
@kylecarbs 
For some reason container feels like a subset of conainter-vm, making it potentially confusing as a negator.

kylecarbs:
Fair. Maybe just `--container` and `--vm` then?

Honestly `--container` and `--cvm` isn't bad either.

kylecarbs:
Whats the news here moogman?

cmoog:
Thought I would just mull it over. The relationship between `--container` and `--cvm` is still confusing to me.

kylecarbs:
definitely fair

fuskovic (approved):

PR #202, opened 2020-12-09
Build linux/windows on ubuntu runner

cmoog:
This PR modifies the build process to use an ubuntu runner for windows/linux binary compilation. This fixes some Windows Defender issues caused by macOS go cross compilation.

### DIFF
--- a/ci/steps/build.sh
+++ b/ci/steps/build.sh
@@ -10,52 +10,38 @@ cd "$(git rev-parse --show-toplevel)/ci/steps"
 
 tag=$(git describe --tags)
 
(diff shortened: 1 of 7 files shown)


ci/steps/build.sh
```diff
@@ -10,52 +10,38 @@ cd "$(git rev-parse --show-toplevel)/ci/steps"
```
deansheather, line 10: Would be good if we got rid of this script entirely in favour of it being inlined in the makefile

deansheather (approved):

PR #208, opened 2020-12-25
docs: add brew install snippet to readme

cmoog:


cmoog:
We just need to make the tap public

cmoog:
I'm considering renaming the Coder Homebrew tap such that the install command reads:
```sh
brew install cdr/tap/coder
```
This is similar to https://github.com/hashicorp/homebrew-tap/blob/master/Formula/terraform-ls.rb

We could also rename the formula itself to `coder-cli` and have an install command of
```sh
brew install cdr/coder/coder-cli
brew install cdr/tap/coder-cli
```

PR #216, opened 2021-01-14
chore: use access url from env resource pool

cmoog:
cc @f0ssel 

[ch5790]

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -15,6 +15,7 @@ import (
 	"cdr.dev/coder-cli/pkg/clog"
 
 	"cdr.dev/coder-cli/coder-sdk"
+	"cdr.dev/coder-cli/internal/coderutil"
 	"cdr.dev/coder-cli/internal/config"
 	"github.com/spf13/cobra"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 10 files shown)


cmoog:
Waiting to test properly when `AccessURL` is available.

cmoog:
Oh there is an issue with the sshAvailable check, that needs to be on a per resource pool basis.

f0ssel:
Looking good so far, hmu when you got the ssh fix in

cmoog:
@f0ssel it's good to go 👍

f0ssel (approved):

internal/cmd/configssh.go
```diff
@@ -103,20 +102,17 @@ func configSSH(configpath *string, remove *bool) func(cmd *cobra.Command, _ []st
...
 
-		if !sshAvailable(envs) {
-			return xerrors.New("SSH is disabled or not available for any environments in your Coder Enterprise deployment.")
-		}
-
-		err = canConnectSSH(ctx)
+		envsWithPools, err := coderutil.EnvsWithPool(ctx, client, envs)
 		if err != nil {
-			return xerrors.Errorf("check if SSH is available: unable to connect to SSH endpoint: %w", err)
+			return xerrors.Errorf("resolve env pools: %w", err)
```
deansheather, line 107: Should we keep canConnectSSH?

deansheather (approved):

deansheather:
oh too late

internal/cmd/configssh.go
cmoog: It was kinda dumb anyway....

PR #219, opened 2021-01-15
bug: fix an issue where query parameters weren't being added to the r…

sreya:
…equest


coder-sdk/request.go
```diff
@@ -43,6 +43,9 @@ func (c Client) request(ctx context.Context, method, path string, in interface{}
 	if err != nil {
 		return nil, xerrors.Errorf("create request: %w", err)
 	}
+	if config.Query != nil {
```
cmoog, line 46: we should do this above `eq, err := http.NewRequestWithContext(`, and add `path` to url.Path`

cmoog (approved):

PR #223, opened 2021-01-21
feat: Prompt user to rebuild workspace on coder sh

Emyrk:
## Quality of Life Feature

When a user does `coder sh <env>`, the cli first checks if that environment is `OFF` or if it requires a rebuild before using. It will then prompt the user if they want to rebuild the workspace.

## To test

**Off**
```
coder env stop <workspace>
coder sh <workspace>
```

**Rebuild Message**

If the environment has a required rebuild message, the output looks like this:
Screenshot from 2021-01-21 16-22-16

## Questions:

Currently `--force` does not work for this feature, and there is no `--no-rebuild` flag. The current command has flag parsing disabled. Are these kinds of flags required? I imagine a user prompt would make scripting `coder sh` rather annoying.

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/time/rate"
@@ -65,9 +66,11 @@ func shValidArgs(cmd *cobra.Command, args []string) error {
 
 func shCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:                "sh [environment_name] [<command [args...]>]",
-		Short:              "Open a shell and execute commands in a Coder environment",
-		Long:               "Execute a remote command on the environment\\nIf no command is specified, the default shell is opened.",
+		Use:   "sh [environment_name] [<command [args...]>]",
+		Short: "Open a shell and execute commands in a Coder environment",
+		Long: `Execute a remote command on the environment
+If no command is specified, the default shell is opened.
+If the command is run in an interactive shell, a user prompt will occur if the environment needs to be rebuilt.`,
 		Args:               shValidArgs,
 		DisableFlagParsing: true,
 		ValidArgsFunction:  getEnvsForCompletion(coder.Me),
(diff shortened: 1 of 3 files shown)


internal/cmd/shell.go
```diff
@@ -90,7 +91,113 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+
+		var rebuildReason string // Either because it is off, or a rebuild is required
+		if env.LatestStat.ContainerStatus == coder.EnvironmentOff {
+			rebuildReason = fmt.Sprintf("Environment %q is currently not running.", env.Name)
+		} else {
+			// Print out the reasons the rebuild is required. Newlines don't work well with
+			// the prompt label, so printing it above the prompt will have to do.
+			clog.LogWarn("rebuild required", requiredRebuilds...)
+			rebuildReason = fmt.Sprintf("Environment %q requires a rebuild to work correctly.", env.Name)
+		}
```
cmoog, line 133: nit: extract to function that returns the `rebuildReason`

internal/cmd/shell.go
```diff
@@ -90,7 +91,113 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+		fallthrough // Fallthrough to watching the logs
+	case env.LatestStat.ContainerStatus == coder.EnvironmentCreating:
+		// Environment is in the process of being created, just trail the logs
+		// and wait until it is done
+		clog.LogInfo(fmt.Sprintf("Environment %q is being rebuilt to be ready for use.", env.Name))
+
+		// Watch the rebuild.
+		// TODO: (@emyrk) Does trailBuildLogs ensure the container is "ON" when it returns?
+		//		Or do I need to poll `findEnv` until it is 'ON'
+		if err = trailBuildLogs(ctx, client, env.ID); err != nil {
```
cmoog, line 164: you're exactly right, `trailBuildLogs` exiting is not indicative of a _successful_ build.

internal/cmd/shell.go
```diff
@@ -90,7 +91,113 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+			return err
+		}
+		// newline after trailBuildLogs to place user on a fresh line for their shell
+		fmt.Println()
+
+		// Find the status of the environment again for use outside the switch statement.
+		env, err = findEnv(ctx, client, envName, coder.Me)
+		if err != nil {
+			return err
+		}
```
cmoog, line 174: Since we already know the ID, let's just fetch by that to avoid the search

internal/cmd/shell.go
```diff
@@ -90,7 +91,113 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+			// the prompt label, so printing it above the prompt will have to do.
+			clog.LogWarn("rebuild required", requiredRebuilds...)
+			rebuildReason = fmt.Sprintf("Environment %q requires a rebuild to work correctly.", env.Name)
+		}
+
+		confirm := promptui.Prompt{
+			Label:     rebuildReason + " Do you wish to rebuild it now? (this will take a moment)",
+			IsConfirm: true,
+		}
+		if _, err := confirm.Run(); err != nil {
```
cmoog, line 139: we should only prompt the user if they are in an interactive terminal

internal/cmd/shell.go
```diff
@@ -90,7 +91,113 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+	if env.LatestStat.ContainerStatus != coder.EnvironmentOn {
+		return clog.Fatal(
+			"the environment is currently offline",
+			fmt.Sprintf("environment %q is not running", env.Name),
+			fmt.Sprintf("its current status is %q", env.LatestStat.ContainerStatus),
+			clog.BlankLine,
+			clog.Tipf("run \"coder envs rebuild %s --follow\" to start the environment", env.Name),
+		)
+	}
+
```
cmoog, line 199: this function definitely needs a reduction in complexity.

cmoog (changes requested):

internal/cmd/shell.go
Emyrk: Added EnvByID function to the client

internal/cmd/shell.go
Emyrk: Factored out a lot of it. I was in the midst of factoring. Launched that draft PR a bit too quick

internal/cmd/shell.go
Emyrk: Added a retry/wait mechanism to check for it, then I re-read this.

When the `trailBuildLogs` finishes, is it possible to still be in `CREATING` state? 
I will check if it's in the `OFF` or `FAILED`. I only need the wait mechanism if it might still be in the `CREATING` state, like a small race condition or something.

internal/cmd/shell.go
Emyrk: Is this what you had in mind to check for that?
https://github.com/cdr/coder-cli/blob/423279755e6607dc1236066d47fd5333d25268b5/internal/cmd/shell.go#L107-L109

So these will **not** prompt:

```
$GOPATH/bin/coder sh tiny | echo
`/home/coder/go/bin/coder sh tiny`
```

Is there a specific way of executing you had in mind? I took this code from the `runCommand`, an alternative is I could use https://rosettacode.org/wiki/Check_output_device_is_a_terminal#Go. I think they do the same thing, so I kept it consistent with the other way we did it.

internal/cmd/shell.go
cmoog: If `trailBuildLogs` completes without an error, I think we can assume it is not `CREATING`.

internal/cmd/shell.go
cmoog: Yeah that's what I had in mind. That said, point taken... this is certainly leaky behavior. Since this is more of nicety, and we can assume 90% of cases will be `coder sh [env]` without pipes or args... I think we should do the cautious thing and only show the prompt when we know for certain the output is interactive. 

Regarding `coder sh tiny | echo`, erroring out as we do now with the helpful error message seems sufficient.

internal/cmd/shell.go
```diff
@@ -65,9 +66,10 @@ func shValidArgs(cmd *cobra.Command, args []string) error {
 
 func shCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:                "sh [environment_name] [<command [args...]>]",
-		Short:              "Open a shell and execute commands in a Coder environment",
-		Long:               "Execute a remote command on the environment\\nIf no command is specified, the default shell is opened.",
+		Use:   "sh [environment_name] [<command [args...]>]",
+		Short: "Open a shell and execute commands in a Coder environment",
+		Long: "Execute a remote command on the environment\\nIf no command is specified, the default shell is opened.\\n" +
```
cmoog, line 71: I think the docs will render properly if we use a real multi-line string here.

internal/cmd/shell.go
Emyrk: I'll remove my polling then.

internal/cmd/shell.go
Emyrk: I left a comment that `--force` or `--no-prompt` would be a clear way for someone to determine behavior, but adding flags to this command is non-trivial because the flag parsing is disabled

internal/cmd/shell.go
Emyrk: It will yes, I can change it to a multiline. I dislike the text tabbed all the way left in my IDE ha.

internal/cmd/shell.go
Emyrk: I am assuming running this via `cron` and other ways won't trigger the prompt too. So I can see how this is helpful, but I don't think it's obvious if I was just writing some bash scripts. Bash scripts will still trigger the prompt if you just run them `./script.sh`.

Emyrk:
@cmoog addressed all comments, the last sticky point is if the current interactive shell behavior is sufficient

internal/cmd/shell.go
```diff
@@ -99,6 +123,132 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+// rebuildPrompt returns function that prompts the user if they wish to
+// rebuild the selected environment if a rebuild is needed. The returned prompt function will
+// return an error if the user selects "no".
+// This functions returns `nil` if there is no reason to prompt the user to rebuild
+// the environment.
+func rebuildPrompt(env *coder.Environment) (prompt func() error) {
+	// Option 1: If the environment is off, the rebuild is needed
+	if env.LatestStat.ContainerStatus == coder.EnvironmentOff {
+		confirm := promptui.Prompt{
+			Label:     fmt.Sprintf("Environment %q is currently not running. Do you wish to rebuild it now? (this will take a moment)", env.Name),
```
cmoog, line 135:
```suggestion
			Label:     fmt.Sprintf("Environment %q is \"OFF\". Rebuild it now? (this can take several minutes)", env.Name),
```

internal/cmd/shell.go
```diff
@@ -99,6 +123,132 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
...
+		// Start the rebuild
+		if err := client.RebuildEnvironment(ctx, env.ID); err != nil {
+			return err
+		}
+
+		fallthrough // Fallthrough to watching the logs
+	case env.LatestStat.ContainerStatus == coder.EnvironmentCreating:
+		// Environment is in the process of being created, just trail the logs
+		// and wait until it is done
+		clog.LogInfo(fmt.Sprintf("Environment %q is in the process of being built.", env.Name))
```
cmoog, line 209:
```suggestion
		clog.LogInfo(fmt.Sprintf("Rebuilding %q", env.Name))
```

internal/cmd/shell.go
cmoog: Yes, agreed. Doesn't appear to be an easy win here.

cmoog (approved):
Looks good. Nice work.

I think the interactive behavior, as you have it, is the best we're going to get for now.

PR #224, opened 2021-01-22
Create login shell with coder sh

jawnsy:
When creating a shell, we should use "exec -a" to set argv[0]
to the basename of the command being run, prefixing a hyphen.
This is how su --login creates login shells: it runs /bin/bash
as "-bash", causing bash to execute .bash_profile, etc.

We change the use of awk to cut, since cut is part of coreutils
and thus more likely to be installed than gawk. We also pass the
uid (from "id -u") instead of username to getent.

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -85,7 +85,9 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
 		args = append(args, strings.Join(cmdArgs[1:], " "))
 	} else {
 		// Bring user into shell if no command is specified.
-		args = append(args, "exec $(getent passwd $(whoami) | awk -F: '{ print $7 }')")
+		shell := "$(getent passwd $(id -u) | cut -d: -f 7)"
+		name := "-$(basename " + shell + ")"
+		args = append(args, fmt.Sprintf("exec -a %q %q", name, shell))
 	}
 
 	envName := cmdArgs[0]


kylecarbs (approved):

internal/cmd/shell.go
```diff
@@ -85,7 +85,9 @@ func shell(cmd *cobra.Command, cmdArgs []string) error {
 		args = append(args, strings.Join(cmdArgs[1:], " "))
 	} else {
 		// Bring user into shell if no command is specified.
-		args = append(args, "exec $(getent passwd $(whoami) | awk -F: '{ print $7 }')")
+		shell := "$(getent passwd $(id -u) | cut -d: -f 7)"
+		name := "-$(basename " + shell + ")"
+		args = append(args, fmt.Sprintf("exec -a %q %q", name, shell))
```
cmoog, line 90: just saying, I've been burned by `%q` before...

cmoog (approved):

internal/cmd/shell.go
cmoog: but I assume that was user error on my part 😆

cmoog:
@jawnsy no need to use a fork. I _think_ you should have the proper perms to push to `cdr/coder-cli`. Let me know if that's not the case.

internal/cmd/shell.go
jawnsy: Haha it might be, but it seems to work alright? https://play.golang.org/p/iHJ_QWLS_Tx

jawnsy:
> @jawnsy no need to use a fork. I _think_ you should have the proper perms to push to a branch in `cdr/coder-cli`. Let me know if that's not the case.

Haha I like the forking workflow, though :grin:

PR #226, opened 2021-01-24
fix: Force bash for non-standard flags to exec

jawnsy:
This addresses a regression introduced by #224. POSIX `sh` does not define arguments to the `exec` built-in, resulting in an error when using `exec -a` to set the process name (`argv[0]`).

This change executes `/bin/bash` instead of `sh`, and also changes to use `exec -l`, which automatically handles the hyphen prefix required to trigger login shell behavior.

This also explicitly runs `/bin/sh`, aligning the behavior of `coder sh` and the frontend Terminal application.

Tested this using Arch Linux (enterprise-dev image) as well as the base Debian image from DockerHub.

```console
[coder@jawnsy-m coder-cli]$ go install ./cmd/coder
[coder@jawnsy-m coder-cli]$ which coder
/home/coder/go/bin/coder
[coder@jawnsy-m coder-cli]$ coder sh jawnsy-m
  | Coder CLI version: unknown
  | Coder API version: 1.14.0+340-gd64a44cb3-20210123
  | 
[coder@jawnsy-m ~]$ shopt login_shell
login_shell     on
[coder@jawnsy-m ~]$ exit
logout
[coder@jawnsy-m coder-cli]$ coder sh jawnsy-debian
  | Coder CLI version: unknown
  | Coder API version: 1.14.0+340-gd64a44cb3-20210123
  | 
root@jawnsy-debian:~# shopt login_shell
login_shell     on
root@jawnsy-debian:~# exit
logout
```

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -82,15 +82,20 @@ coder sh front-end-dev cat ~/config.json`,
 
 func shell(cmd *cobra.Command, cmdArgs []string) error {
 	ctx := cmd.Context()
-	command := "sh"
-	args := []string{"-c"}
+	var command string
+	var args []string
 	if len(cmdArgs) > 1 {
+		command = "/bin/sh"
+		args = []string{"-c"}
 		args = append(args, strings.Join(cmdArgs[1:], " "))
 	} else {
 		// Bring user into shell if no command is specified.
 		shell := "$(getent passwd $(id -u) | cut -d: -f 7)"
-		name := "-$(basename " + shell + ")"
-		args = append(args, fmt.Sprintf("exec -a %q %q", name, shell))
+
+		// force bash for the '-l' flag to the exec built-in
+		command = "/bin/bash"
+		args = []string{"-c"}
+		args = append(args, fmt.Sprintf("exec -l %q", shell))
 	}
 
 	envName := cmdArgs[0]


cmoog (approved):

internal/cmd/shell.go
```diff
@@ -82,15 +82,20 @@ coder sh front-end-dev cat ~/config.json`,
...
 	} else {
 		// Bring user into shell if no command is specified.
 		shell := "$(getent passwd $(id -u) | cut -d: -f 7)"
-		name := "-$(basename " + shell + ")"
-		args = append(args, fmt.Sprintf("exec -a %q %q", name, shell))
+
+		// force bash for the '-l' flag to the exec built-in
+		command = "/bin/bash"
+		args = []string{"-c"}
+		args = append(args, fmt.Sprintf("exec -l %q", shell))
```
tychoish, line 98: I don't instinctively trust %q to handle shell quoting rules, but this seems like it's good.

tychoish (approved):

internal/cmd/shell.go
jawnsy: hahaha @cmoog was suspicious of this too, perhaps I should change it? the Terminal app in the frontend is not quoted, and it works fine

PR #227, opened 2021-01-24
fix: Escape shell arguments

jawnsy:
The commands passed to "coder sh" are passed as a single argument to "sh -c", so we need to shell-escape the command we pass. This change escapes spaces, backslash, and quotes.

I'm not 100% sure if this is a good idea, since it is a breaking change in case people have already been escaping their commands.

Note: this will need to be rebased once #226 is merged

Tested as follows:

```console
[coder@jawnsy-m coder-cli]$ go install ./cmd/coder
[coder@jawnsy-m coder-cli]$ which coder
/home/coder/go/bin/coder
[coder@jawnsy-m coder-cli]$ coder sh jawnsy-m go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
  | Coder CLI version: unknown
  | Coder API version: 1.14.0+340-gd64a44cb3-20210123
  | 
argv[0] = /tmp/go-build627553793/b001/exe/test
argv[1] = 1
argv[2] = 2
argv[3] = 3 4
argv[4] = "abc def" \\abc
argv[5] = 5
argv[6] = 6
argv[7] = 7 8 9
[coder@jawnsy-m coder-cli]$ go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
argv[0] = /tmp/go-build909559834/b001/exe/test
argv[1] = 1
argv[2] = 2
argv[3] = 3 4
argv[4] = "abc def" \\abc
argv[5] = 5
argv[6] = 6
argv[7] = 7 8 9
```

For comparison, this is what `sh -c` and `exec` would look like:

```console
[coder@jawnsy-m coder-cli]$ sh -c "go run ~/test.go 1 2 \"3 4\" '\"abc def\" \\abc' 5 6 \"7 8 9\""
argv[0] = /tmp/go-build754337722/b001/exe/test
argv[1] = 1
argv[2] = 2
argv[3] = 3 4
argv[4] = "abc def" \abc
argv[5] = 5
argv[6] = 6
argv[7] = 7 8 9
[coder@jawnsy-m coder-cli]$ bash
[coder@jawnsy-m coder-cli]$ exec go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
argv[0] = /tmp/go-build905411501/b001/exe/test
argv[1] = 1
argv[2] = 2
argv[3] = 3 4
argv[4] = "abc def" \\abc
argv[5] = 5
argv[6] = 6
argv[7] = 7 8 9
```

And this was the previous behavior:

```console
[coder@jawnsy-m coder-cli]$ which coder
/opt/coder/coder-cli/coder
[coder@jawnsy-m coder-cli]$ coder sh jawnsy-m go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
  | Coder CLI version: v1.15.0-8-gae35620
  | Coder API version: 1.14.0+340-gd64a44cb3-20210123
  | 
argv[0] = /tmp/go-build394408240/b001/exe/test
argv[1] = 1
argv[2] = 2
argv[3] = 3
argv[4] = 4
argv[5] = abc def
argv[6] = \abc
argv[7] = 5
argv[8] = 6
argv[9] = 7
argv[10] = 8
argv[11] = 9
```

### DIFF
--- a/internal/cmd/shell.go
+++ b/internal/cmd/shell.go
@@ -81,14 +81,46 @@ coder sh front-end-dev cat ~/config.json`,
 	}
 }
 
+// shellEscape escapes an argument so that we can pass it 'sh -c'
+// and have it do the right thing.
+//
+// Use this to ensure that the result of a command running in
+// the development environment behaves the same as the command
+// running via "coder sh".
+//
+// For example:
+//
+//   $ coder sh env
+//   $ go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
+//
+// should produce the same output as:
+//
+//   $ coder sh go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
+func shellEscape(arg string) string {
+	r := strings.NewReplacer(`\`, `\\`, `"`, `\"`, `'`, `\'`, ` `, `\ `)
+	return r.Replace(arg)
+}
+
 func shell(cmd *cobra.Command, cmdArgs []string) error {
 	ctx := cmd.Context()
+
 	var command string
 	var args []string
 	if len(cmdArgs) > 1 {
+		var escapedArgs strings.Builder
+
+		for i, arg := range cmdArgs[1:] {
+			escapedArgs.WriteString(shellEscape(arg))
+
+			// Add spaces between arguments, except the last argument
+			if i < len(cmdArgs)-2 {
+				escapedArgs.WriteByte(' ')
+			}
+		}
+
 		command = "/bin/sh"
 		args = []string{"-c"}
-		args = append(args, strings.Join(cmdArgs[1:], " "))
+		args = append(args, escapedArgs.String())
 	} else {
 		// Bring user into shell if no command is specified.
 		shell := "$(getent passwd $(id -u) | cut -d: -f 7)"
--- a/internal/cmd/shell_test.go
+++ b/internal/cmd/shell_test.go
@@ -1,0 +1,41 @@
+package cmd
+
+import "testing"
+
+func TestShellEscape(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		Name    string
+		Input   string
+		Escaped string
+	}{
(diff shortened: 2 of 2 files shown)


internal/cmd/shell.go
```diff
@@ -80,17 +80,54 @@ coder sh front-end-dev cat ~/config.json`,
...
+//
+//   $ coder sh env
+//   $ go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
+//
+// should produce the same output as:
+//
+//   $ coder sh go run ~/test.go 1 2 "3 4" '"abc def" \\abc' 5 6 "7 8 9"
+func shellEscape(arg string) string {
+	r := strings.NewReplacer(`\`, `\\`, `"`, `\"`, `'`, `\'`, ` `, `\ `)
+	return r.Replace(arg)
```
tychoish, line 100: I think eventually we should be relying on shlex (or similar) to be parsing and escaping shell commands from strings into arg arrays, and avoid rolling our own for things like this.

tychoish (approved):

internal/cmd/shell.go
jawnsy: @tychoish Agree, I did a quick search bit didn't find any good ones. Happy to switch over now, or we can do it later, WDYT?

tychoish: https://github.com/google/shlex is functional. 

I think this might also solve the %q question in the other PR

jawnsy: @tychoish Oh yeah, that looks much better. Looks like there were a few cases that I missed.

I don't think this will work for #226 because we need the shell to interpret the subshell calls for getent/etc, but I will give it a try

jawnsy: Hmm it looks like shlex is for parsing shell expressions, how do I use it to escape things?

cmoog (approved):

PR #230, opened 2021-01-25
chore: Clean up CI scripts

jawnsy:
* Build with Go provided by GitHub instead of building inside a Dockerfile, which should speed up the build
* Condense build steps to avoid repeatedly downloading cached Go packages, which should speed up the build
* Run goimports using golangci-lint, rather than running gofmt and goimports manually in make fmt
* Skip signing/notarizing Darwin binary in build, since this takes a long time and is done in release
* Only build/test for pull requests against master
* Remove unnecessary steps in ci/image/Dockerfile
  * goimports is managed by golangci-lint
  * grep is already installed in the base image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,11 @@
 name: build
-on: [push]
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


jawnsy:
@cmoog Some small cleanups to the build here, but I'm not really sure if it's an improvement. What do you think?

cmoog:
I could go either way with these changes. I would still like to run the notarized build on master to ensure we don't get surprised with errors when releases are tagged.

.github/workflows/test.yaml
```diff
@@ -12,20 +16,30 @@ jobs:
...
+        run: make gendocs
+      - name: Check that "make gendocs" is clean
+        run: ./ci/scripts/files_changed.sh
+
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: golangci-lint
+      - uses: actions/checkout@v1
```
tychoish, line 36: my moving from 2 to one?

.github/workflows/build.yaml
```diff
@@ -19,20 +23,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v1
-      - name: Install Gon
-        run: |
-          brew tap mitchellh/gon
-          brew install mitchellh/gon/gon
-      - name: Import Signing Certificates
-        uses: Apple-Actions/import-codesign-certs@v1
```
tychoish, line 27: it seems like we wouldn't want to drop code signing?

jawnsy:
The useful parts of this change were merged in #243, obsoleting this PR

PR #231, opened 2021-01-26
chore: update template request field name from 'name' to 'file_name'

sreya:


cmoog (approved):
Is `FilePath` more appropriate?

sreya:
I think we're constraining files to the `.coder` directory, which we'll assume is in the root of the repo

cmoog:
Ahhh, makes sense... much better.

PR #236, opened 2021-02-04
feat(cli): AutoStart education notice [ch7441]

vapurrmaid:
I tried my best to get some cool info/ed information around AutoStart into the CLI :zap:

### Env Pre-Run Cmds

I added this to:

- envs create
- envs rebuild

### Env Create

```
$ go run ./cmd/coder/main.go envs create created-from-cli --image ubuntu --enable-autostart --org Coder
info: ⚡NEW: Automate daily environment startup
  | Visit https://master.cdr.dev/preferences?tab=autostart to configure your preferred time
  | Coder CLI version: unknown
  | Coder API version: 1.14.0+743-g7c24d16bb-20210216
  | 
success: creating environment...
  | 
  | tip: run "coder envs watch-build created-from-cli" to trail the build logs
```

## Next Steps

Should we accept this effort, we should spawn a ticket to remove the education notice from the next non-patch release

### DIFF
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -152,16 +152,16 @@ coder envs --user charlie@coder.com ls -o json \
 
 func createEnvCmd() *cobra.Command {
 	var (
-		org    string
-		cpu    float32
-		memory float32
-		disk   int
-		gpus   int
-		img    string
-		tag    string
-		follow bool
-		useCVM bool
-		useAS  bool
+		org             string
+		cpu             float32
+		memory          float32
+		disk            int
+		gpus            int
+		img             string
+		tag             string
+		follow          bool
+		useCVM          bool
+		enableAutostart bool
 	)
 
 	cmd := &cobra.Command{
@@ -214,7 +214,7 @@ coder envs create my-new-powerful-env --cpu 12 --disk 100 --memory 16 --image ub
 				DiskGB:          disk,
 				GPUs:            gpus,
 				UseContainerVM:  useCVM,
-				EnableAutoStart: useAS,
+				EnableAutoStart: enableAutostart,
 			}
 
 			// if any of these defaulted to their zero value we provision
@@ -258,7 +258,7 @@ coder envs create my-new-powerful-env --cpu 12 --disk 100 --memory 16 --image ub
 	cmd.Flags().StringVarP(&img, "image", "i", "", "name of the image to base the environment off of.")
 	cmd.Flags().BoolVar(&follow, "follow", false, "follow buildlog after initiating rebuild")
 	cmd.Flags().BoolVar(&useCVM, "container-based-vm", false, "deploy the environment as a Container-based VM")
-	cmd.Flags().BoolVar(&useAS, "enable-autostart", false, "automatically start this environment at your preferred time.")
+	cmd.Flags().BoolVar(&enableAutostart, "enable-autostart", false, "automatically start this environment at your preferred time.")
 	_ = cmd.MarkFlagRequired("image")
 	return cmd
 }
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -622,7 +622,7 @@ func buildUpdateReq(ctx context.Context, client *coder.Client, conf updateConf) 
 	return &updateReq, nil
 }
 
-// TODO (Grey): Remove education in a future non-patch release
+// TODO (Grey): Remove education in a future non-patch release.
 func autoStartInfo() {
 	var preferencesURI string
 
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -11,6 +11,7 @@ import (
 	"os"
 
 	"cdr.dev/coder-cli/coder-sdk"
+	"cdr.dev/coder-cli/internal/config"
 	"cdr.dev/coder-cli/internal/x/xcobra"
 	"cdr.dev/coder-cli/pkg/clog"
 	"cdr.dev/coder-cli/pkg/tablewriter"
@@ -160,6 +161,7 @@ func createEnvCmd() *cobra.Command {
 		tag    string
 		follow bool
 		useCVM bool
+		useAS  bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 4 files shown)


internal/cmd/autostart.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return cmd
+}
+
+func setAutostartCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set [time]",
+		Short: "set autostart settings owned by the active user",
+		Args:  xcobra.ExactArgs(1),
+		Long:  "Set Coder autostart settings owned by the active user. The suppled time is in HH:mm:ss format and assumes UTC.",
```
vapurrmaid, line 73:
```suggestion
		Long:  "Set Coder autostart settings owned by the active user. The supplied time is in HH:mm:ss format and assumes UTC.",
```

internal/cmd/autostart.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+
+			as, err := client.MyAutoStart(ctx)
+			if err != nil {
+				return err
+			}
+
+			switch outputFmt {
+			case humanOutput:
+				fmt.Printf("AutoStart Time:\t%v\n", as.AutoStart)
```
cmoog, line 50: You _could_ use `tablewriter` here if you wanted to.

internal/cmd/autostart.go
```diff
@@ -0,0 +1,99 @@
...
+			t, err := time.Parse("15:04:05", args[0])
+			if err != nil {
+				return err
+			}
+			updateReq := &coder.AutoStartRequest{
+				AutoStart: t.Format(time.RFC3339),
+			}
+			_, err = client.UpdateAutoStart(ctx, *updateReq)
+			if err != nil {
+				return xerrors.Errorf("update autostart: %w", err)
```
cmoog, line 91: how do we pick the timezone here?

internal/cmd/envs.go
```diff
@@ -57,6 +57,9 @@ func lsEnvsCommand() *cobra.Command {
 		Use:   "ls",
 		Short: "list all environments owned by the active user",
 		Long:  "List all Coder environments owned by the active user.",
+		PreRun: func(cmd *cobra.Command, args []string) {
+			autoStartInfo()
```
cmoog, line 61: I think this could get a little spammy for `envs ls`... maybe we write to a file and don't show the suggested > 1 time? or we only show it for `envs rebuild`?

internal/cmd/envs.go
```diff
@@ -611,3 +626,7 @@ func buildUpdateReq(ctx context.Context, client *coder.Client, conf updateConf)
 	}
 	return &updateReq, nil
 }
+
+func autoStartInfo() {
+	clog.LogInfo("⚡NEW: Automate daily environment startup", "Visit /preferences to configure your preferred time")
```
cmoog, line 631: Since we know the access url, we _could_ link directly to the dash here.

cmoog (changes requested):

internal/cmd/envs.go
cmoog: this value is accessible with the `internal/config` package.

internal/cmd/autostart.go
vapurrmaid: https://github.com/cdr/coder-cli/pull/236/commits/bcb983046de2720621de4ad21f1c8da05305c0f7 I think that's the intended use case?

internal/cmd/autostart.go
vapurrmaid: https://github.com/cdr/coder-cli/pull/236/commits/bcb983046de2720621de4ad21f1c8da05305c0f7 I think that's the intended way to do it here (unsure)?

cmoog: was thinking you could make a pseudo slice with one element, but it doesn't matter either way. whatever you think is cleanest.

internal/cmd/envs.go
vapurrmaid: Thanks! Added here: https://github.com/cdr/coder-cli/pull/236/commits/3cffbca09546735cd0372df8330a3d956d611374

internal/cmd/autostart.go
vapurrmaid: Yea, so I'm a little unfamiliar with some of the tools available to me, but I thought as a starting point we could set the limitation that it's parsed in UTC. The backend expects/uses UTC, and the web client just does that conversion from locale -> UTC. I'm less sure how to do that in a CLI environment.

> In the absence of a time zone indicator, Parse returns a time in UTC.

The next thought was great, maybe we could add a format string that allows an optional timezone, but I wasn't sure how to implement it properly (was referring to these standard layouts: https://yourbasic.org/golang/format-parse-string-time-date-example/

I wanted to avoid the user having to include a date, since the date itself is meaningless: we only care about the time.

If there's a clear way to do this, I'd think ideally we either have a flexible layout string or a flag for the timezone

```
coder autostart set 15:30:00 --timezone=[timezone]
```


Definitely would appreciate help on this one from someone who's more familiar with this side of things. If not, I don't think it's a show stopper to specify the time expects UTC for now

vapurrmaid: I wonder if we can use `--timezone` flag and default to UTC?

cmoog: Honestly I'm not apposed to keeping the actual setting of the auto-start time to dashboard-only. Then we keep the boolean flag for env creating in the CLI.

vapurrmaid: Fair, I'll remove this

internal/cmd/envs.go
```diff
@@ -160,6 +161,7 @@ func createEnvCmd() *cobra.Command {
 		tag    string
 		follow bool
 		useCVM bool
+		useAS  bool
```
scsmithr, line 164: Imo this should be `enableAutostart`. Not clear to me what `useAS` is.

scsmithr (commented):
I'm not concerned about this being spammy. If we plan on removing this after some release, we should document that somewhere.

scsmithr:
I also see you have the "document somewhere" already figured out... didn't read Next Steps.

internal/cmd/envs.go
vapurrmaid: IDK how to idiomatic GO...happy to use the TS convention :laughing:

vapurrmaid: What's an `rp`, what's `u`...IDK

scsmithr: reverse proxy and user, obvs...

cmoog (approved):

PR #238, opened 2021-02-05
Add commands for workspace providers

f0ssel:
This adds create, list, and delete functionality of workspace providers to coder-cli. The functionality is bare bones in this iteration and will be augmented with more documentation, especially around the provider creation flow, in a future pr. 

```
➜  coder git:(cli-wp) ✗ go run main.go providers create test-2
ID                                   Name      Status    EnvproxyToken                        
601dc1e2-be83c6116b11dc4507db5f21    test-2              601dc1e2-be4362804a91d8cb4450ec21  

➜  coder git:(cli-wp) ✗ go run main.go providers ls           
Name      Status     Cluster Address                                     Namespace    Storage Class    Cluster Domain Suffix    Access URL                Devurl Host    SSH Enabled    Namespace Allowlist    
local     ready      https://kubernetes.default.svc.cluster.local:443    f0ssel                        .svc.cluster.local       http://104.154.100.117                   true           []                     
test-1    pending                                                                                                                                                        false          []                     

➜  coder-cli git:(cli-wp) ✗ go run cmd/coder/main.go providers rm test-1 test-2
error: failed to remove workspace provider "test-2"
  | cause: no workspace provider found by name "test-2"

success: removed workspace provider with name "test-1"
fatal: 1 failure emitted

exit status 1
```

Status field is empty on the create command until I push my changes in m to add that to the create response. 

These commands are hidden for now.


coder-sdk/workspace_providers.go
```diff
@@ -0,0 +1,80 @@
...
+	Local               bool                    `json:"local"`
+	ClusterAddress      string                  `json:"cluster_address"`
+	DefaultNamespace    string                  `json:"default_namespace"`
+	StorageClass        string                  `json:"storage_class"`
+	ClusterDomainSuffix string                  `json:"cluster_domain_suffix"`
+	EnvproxyAccessURL   string                  `json:"envproxy_access_url" validate:"required"`
+	DevurlHost          string                  `json:"devurl_host"`
+	SSHEnabled          bool                    `json:"ssh_enabled"`
+	NamespaceWhitelist  []string                `json:"namespace_whitelist"`
+	OrgWhitelist        []string                `json:"org_whitelist"`
```
cmoog, line 22: These should all have `table` fields, with most hidden via `-`.

internal/cmd/providers.go
```diff
@@ -0,0 +1,130 @@
...
+
+			err = client.DeleteWorkspaceProviderByID(ctx, id)
+			if err != nil {
+				return clog.Error(
+					fmt.Sprintf(`failed to delete workspace provider "%s"`, id),
+					clog.Causef(err.Error()),
+				)
+			}
+
+			clog.LogSuccess(fmt.Sprintf(`deleted workspace provider with id "%s"`, id))
```
cmoog, line 124: the name might provider a better UX here.

internal/cmd/providers.go
```diff
@@ -0,0 +1,130 @@
...
+	return cmd
+}
+
+func deleteProviderCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rm [workspace_provider_id]",
+		Short: "remove a workspace provider.",
+		Args:  xcobra.ExactArgs(1),
+		Long:  "Remove an existing Coder workspace provider.",
+		Example: `# remove a new workspace provider by id
```
cmoog, line 106: might not be necessary here, but I like the pattern we have of `rm` accepting > 1 items and removing all.

cmoog (changes requested):

coder-sdk/workspace_providers.go
f0ssel: They displayed nicely in my test, what does the table tags do exactly?

cmoog: If the table tag is not present, we default to the Go identifier name... which isn't exactly good behavior. Plus, the line length of the test shown above it far too large... we should be hiding `ID` and a few others.

internal/cmd/providers.go
f0ssel: Good call, I'm gonna change the entire command to just take a name, so it matches create UX

coder-sdk/workspace_providers.go
```diff
@@ -0,0 +1,97 @@
...
+	var res CreateWorkspaceProviderRes
+	err := c.requestBody(ctx, http.MethodPost, "/api/private/resource-pools", req, &res)
+	if err != nil {
+		return nil, err
+	}
+	return &res, nil
+}
+
+// DeleteWorkspaceProviderByName deletes a workspace provider entity from the Coder control plane.
+func (c *Client) DeleteWorkspaceProviderByName(ctx context.Context, name string) error {
```
cmoog, line 80: so for `coder-sdk` we should only be using the unique identifiers for our base methods like this.

cmoog: something like this belongs in `coderutil` in my view.

internal/cmd/providers.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				return err
+			}
+
+			wps, err := client.WorkspaceProviders(ctx)
+			if err != nil {
+				return xerrors.Errorf("listing workspace providers: %w", err)
+			}
+
+			egroup := clog.LoggedErrGroup()
```
cmoog, line 120: ❤️

cmoog (approved):

PR #239, opened 2021-02-06
ent: Add utf charset to doc and add svg based power button - [ch7623]

kvnlnt:


internal/loginsrv/server.go
```diff
@@ -15,6 +15,7 @@ type Server struct {
 func loginRedirectHTMLDoc(message string, status string) string {
 	htmlTemplate := `<html>
 		<head>
+		<meta http-equiv="Content-Type" content="text/html;charset=UTF-8">
```
vapurrmaid, line 18: I don't have a rhyme or reason, other than it's directly from here: https://www.w3.org/International/questions/qa-html-encoding-declarations

```suggestion
		<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
```

vapurrmaid (approved):

internal/loginsrv/server.go
kvnlnt: good enough reason for me, never know what browsers might trip over.

PR #242, opened 2021-02-12
feat: unconditionally show login URL

jawnsy:
Do not attempt to open the user's browser, because their default
browser may not be their preferred one. The user may be logged in
to Coder using multiple accounts on different browsers, or they
may not use their default browser for Coder.

Tools like gcloud unconditionally show a login URL, so that the
user can decide where to open it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
 	github.com/klauspost/compress v1.10.8 // indirect
 	github.com/manifoldco/promptui v0.7.0
 	github.com/mattn/go-colorable v0.1.8 // indirect
-	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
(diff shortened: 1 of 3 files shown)


tychoish (approved):

cmoog (commented):
I'm honestly not sure whether this is an improvement, but matching the behavior of other tools seems reasonable in this case I suppose.

jawnsy:
@cmoog Alternative approaches that would also work for my use case:

* Do the existing behavior (open browser), show the message only on error
* Do the existing behavior (open browser), show the URL message unconditionally

If you prefer to keep the existing behavior of opening a browser window (personally I find it annoying and a bit surprising, since I didn't explicitly ask for it, but YMMV), I think the second option would work too. Thoughts?

cmoog:
This isn't really applicable anymore given the recent rewrite of our login flow.

PR #244, opened 2021-02-12
test: remove config_test and enable sdk tests

jawnsy:
* Remove config_test, which is not runnable in CI
* Include coder-sdk for unit tests (future)


tychoish (approved):

jawnsy:
@cmoog Seems this test fails, maybe due to permissions of the credentials during integration tests. Should we remove this test entirely?

cmoog:
Sounds good, let's just remove those tests entirely.

PR #245, opened 2021-02-17
feat: configurable context root support

jawnsy:
Append, rather than overwrite, request paths, so that we can invoke
requests on custom context roots.

### DIFF
--- a/coder-sdk/request.go
+++ b/coder-sdk/request.go
@@ -66,8 +66,7 @@ func (c Client) request(ctx context.Context, method, path string, in interface{}
 	if config.Query != nil {
 		url.RawQuery = config.Query.Encode()
 	}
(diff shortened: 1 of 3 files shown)


coder-sdk/request.go
```diff
@@ -65,7 +66,8 @@ func (c Client) request(ctx context.Context, method, path string, in interface{}
 	if config.Query != nil {
 		url.RawQuery = config.Query.Encode()
 	}
-	url.Path = path
+	url.Path = strings.TrimSuffix(url.Path, "/")
+	url.Path = url.Path + path
```
tychoish, line 70: use fmt.Sprint rather than string concatination

jawnsy: done

cmoog (approved):

PR #248, opened 2021-02-19
feat: reuse http.Client and validate client options at creation time

jawnsy:
* Create Client interface, so that it can be mocked/stubbed
* Validate client options at creation time
* Reuse internal HTTP Client, so that we have connection pooling
* Send the Session-Token in a header rather than cookie

### DIFF
--- a/coder-sdk/client.go
+++ b/coder-sdk/client.go
@@ -1,46 +1,265 @@
 package coder
 
 import (
+	"context"
+	"errors"
 	"net/http"
-	"net/http/cookiejar"
 	"net/url"
+
+	"cdr.dev/wsep"
+	"nhooyr.io/websocket"
 )
 
 // Me is the user ID of the authenticated user.
 const Me = "me"
 
 // Client wraps the Coder HTTP API.
-type Client struct {
-	BaseURL *url.URL
-	Token   string
+type Client interface {
+	// PushActivity pushes CLI activity to Coder.
+	PushActivity(ctx context.Context, source, envID string) error
+
+	// Me gets the details of the authenticated user.
+	Me(ctx context.Context) (*User, error)
+
+	// UserByID get the details of a user by their id.
+	UserByID(ctx context.Context, id string) (*User, error)
+
+	// SSHKey gets the current SSH kepair of the authenticated user.
+	SSHKey(ctx context.Context) (*SSHKey, error)
+
+	// Users gets the list of user accounts.
+	Users(ctx context.Context) ([]User, error)
+
+	// UserByEmail gets a user by email.
+	UserByEmail(ctx context.Context, email string) (*User, error)
+
+	// UpdateUser applyes the partial update to the given user.
+	UpdateUser(ctx context.Context, userID string, req UpdateUserReq) error
+
+	// UpdateUXState applies a partial update of the user's UX State.
+	UpdateUXState(ctx context.Context, userID string, uxsPartial map[string]interface{}) error
+
+	// CreateUser creates a new user account.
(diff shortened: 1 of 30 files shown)


coder-sdk/client.go
```diff
@@ -1,46 +1,265 @@
...
+
+	// EditEnvironment modifies the environment specification and initiates a rebuild.
+	EditEnvironment(ctx context.Context, envID string, req UpdateEnvironmentReq) error
+
+	// DialWsep dials an environments command execution interface
+	// See https://github.com/cdr/wsep for details.
+	DialWsep(ctx context.Context, baseURL *url.URL, envID string) (*websocket.Conn, error)
+
+	// DialExecutor gives a remote execution interface for performing commands inside an environment.
+	DialExecutor(ctx context.Context, baseURL *url.URL, envID string) (wsep.Execer, error)
```
tychoish, line 114: is wsep.Execer still available since the repo was archived? 

url.URL also seems like an interesting API choice.

tychoish (approved):
love it.

coder-sdk/client.go
jawnsy: I dunno, I didn't really look at any of these APIs, I just extracted the public methods into an interface. Is there something we should switch over to?

coder-sdk/client.go
```diff
@@ -1,46 +1,265 @@
...
-// WARNING:
-// If the caller sets a custom transport to set TLS settings or a custom CA the default
-// pool will not be used and it might result in a new dns lookup/tls session/socket begin
-// established each time.
-func (c Client) newHTTPClient() (*http.Client, error) {
-	jar, err := cookiejar.New(nil)
-	if err != nil {
-		return nil, err
+// NewClient creates a new default Coder SDK client.
+func NewClient(opts ClientOptions) (c Client, err error) {
```
jawnsy, line 236: @cmoog What's your opinion here? I know a common practice is to [return the struct rather than an interface](https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8), which would make this change less invasive. My goal here was to force the use of a constructor, so that we can better control mutable vs immutable state, and so that we can handle alternative construction methods (e.g. create a Client with either an API Token or a username/password combination).

If we return the struct, then we should make it a public type so that it can be assigned (e.g. `var x coder.DefaultClient = abc`), but this means users have to make sure to do that instead of creating an empty Client (e.g. `client := &coder.DefaultClient{}`) which would be useless as it'd be missing the necessary baseURL etc settings

coder-sdk/client.go
cmoog: ah whoops, probably need to unarchive the wsep repo for this reason.

coder-sdk/client.go
cmoog: I'm not too opinionated here, so I'd generally want us to optimize for the simplest, lowest maintenance path. As far as matching other libraries, as you pointed out with Kubernetes (and [AWS](https://github.com/aws/aws-sdk-go/blob/master/aws/client/client.go) and [GCP](https://github.com/googleapis/google-cloud-go/blob/master/spanner/client.go#L78:6))  it is very standard for packages to expose a public struct type for accessing an underlying HTTP API.

cmoog (approved):
I could go either way on this one, so your call. One additional data point for consideration, take a look at how AWS does it with a separate package for the interface definitions https://github.com/aws/aws-sdk-go/blob/master/service/ec2/ec2iface/interface.go.
 Not saying this is particularly clean, just another option.

PR #250, opened 2021-02-23
feat: add password authentication to coder-sdk

jawnsy:
This enables authenticating to Coder using the coder-sdk using an
email/password combination (basic authentication). As it is intended
for test purposes only, it is not exposed via coder-cli.

### DIFF
--- a/coder-sdk/client.go
+++ b/coder-sdk/client.go
@@ -29,26 +29,26 @@ type ClientOptions struct {
 	// Token is the API Token used to authenticate (optional).
 	//
 	// If Token is provided, the DefaultClient will use it to
-	// authenticate. If it is not provided, it will require another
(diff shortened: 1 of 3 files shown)


coder-sdk/client.go
```diff
@@ -14,15 +18,38 @@ const Me = "me"
...
+	Email string
+
+	// Password will be used to authenticate.
+	//
+	// DefaultClient requires exactly one type of credential
+	// (Token, or Email/Password pair).
+	//
+	// NewClient will exchange the Email and Password for a Token
+	// and will not retain them in-memory after NewClient returns.
+	Password string
```
deansheather, line 52: Should document here that email and password is only supported for built in auth, not oidc.

deansheather (approved):

deansheather (approved):

PR #251, opened 2021-02-24
feat: enable password changes in coder-sdk

jawnsy:
* Allow modification of passwords and the temporary_password flag for users managed by the built-in authentication provider.
* Split unit tests into separate files corresponding to the implementations.


coder-sdk/test/login_cmd.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"cdr.dev/coder-cli/coder-sdk"
+)
+
+func main() {
```
deansheather, line 11: Why is this a main package if it's for testing?

coder-sdk/users_test.go
```diff
@@ -0,0 +1,104 @@
...
+		HTTPClient: server.Client(),
+		Token:      "JcmErkJjju-KSrztst0IJX7xGJhKQPtfv",
+	})
+	require.NoError(t, err, "failed to create coder.Client")
+
+	err = client.UpdateUser(context.Background(), "me", coder.UpdateUserReq{
+		UserPasswordSettings: &coder.UserPasswordSettings{
+			OldPassword: "vt9g9rxsptrq",
+			Password:    "wmf39jw2f7pk",
+			Temporary:   true,
```
deansheather, line 100: If you're changing your own password you're not intended to set temporary = true, the temporary flag is for when accounts are created or password is changed by an admin.

deansheather (approved):

coder-sdk/test/login_cmd.go
jawnsy: oops, did not mean to check this in

coder-sdk/users_test.go
jawnsy: fair enough, made this false

PR #252, opened 2021-02-25
chore: expose API Token and BaseURL

jawnsy:
This adds accessors to retrieve the Client's active Session Token
and Base URL.

### DIFF
--- a/coder-sdk/client_test.go
+++ b/coder-sdk/client_test.go
@@ -18,6 +18,7 @@ func TestAuthentication(t *testing.T) {
 	t.Parallel()
 
 	const token = "g4mtIPUaKt-pPl9Q0xmgKs7acSypHt4Jf"
+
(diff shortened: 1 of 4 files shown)


coder-sdk/client_test.go
```diff
@@ -103,6 +106,7 @@ func TestPasswordAuthentication(t *testing.T) {
 		Password:   "coder4all",
 	})
 	require.NoError(t, err, "failed to create Client")
+	require.Equal(t, "g4mtIPUaKt-pPl9Q0xmgKs7acSypHt4Jf", client.Token(), "expected token to match")
```
deansheather, line 109: Should store this token in a variable or constant.

deansheather (approved):

coder-sdk/client_test.go
jawnsy: I kinda went back and forth on it, using constants could obscure test problems and also obscure what the tests are testing, but I don't feel strongly about it so will switch over to constants

cmoog (approved):

PR #258, opened 2021-03-05
chore: provide resource pool id and namespace in env create

cmoog:
We can expose these values via a flag later, but for now using the default pool is OK.

### DIFF
--- a/internal/coderutil/env.go
+++ b/internal/coderutil/env.go
@@ -59,7 +59,7 @@ func EnvsWithProvider(ctx context.Context, client coder.Client, envs []coder.Env
 	return pooledEnvs, nil
 }
 
-// DefaultWorkspaceProvider returns the default provider with with to create environments.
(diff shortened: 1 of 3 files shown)


internal/coderutil/env.go
```diff
@@ -58,3 +58,17 @@ func EnvsWithProvider(ctx context.Context, client coder.Client, envs []coder.Env
...
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range provider {
+		if p.Local {
+			return &p, nil
+		}
+	}
+	return nil, coder.ErrNotFound
+}
```
cmoog, line 74: cc @f0ssel

f0ssel (approved):
looks good for now, when we'll do the local -> default migration we can also consider whether we can rip out the `Local` bool altogether as well, but that's down the road.

Emyrk (commented):
I know it's already merged, but :+1:

Emyrk:
I know it's already merged, but :+1:

PR #259, opened 2021-03-06
chore: use dependabot to manage dependencies

jawnsy:
Use dependabot to manage the dependencies defined in go.mod and
GitHub Actions workflows, so that we can proactively update versions.

Outdated versions of third-party dependencies frequently have known
security vulnerabilities with CVEs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"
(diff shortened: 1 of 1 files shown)


deansheather (approved):
Do we need to get a github admin to enable depdendabot on this repo for this to work?

cmoog (approved):

jawnsy:
@deansheather That is an excellent point, and would explain why there wasn't a dependency check :joy:

cmoog:
Should be good to go now.

PR #271, opened 2021-03-08
Handle wac template parse errors

Emyrk:
Related to: https://github.com/cdr/m/pull/7373

In addition to the errors reported in the `/m` pr, this also now expands verbose errors. These verbose errors are not yet reviewed for human readability:

## 404 Repo

```
error: unexpected status code 500: An internal server error occurred.
  | read file: get file contents: GET https://api.github.com/repos/Emyrk/te/contents/.coder/coder.yaml?ref=master: 404 Not Found []
```

## Empty Request

```
error: unexpected status code 500: An internal server error occurred.
  | parse repo url: unexpected url "", expecting at least an owner and a repo
```

## Any server internal error

```
error: unexpected status code 500: An internal server error occurred.
  | exec select: pq: relation "images" does not exist
```

### DIFF
--- a/coder-sdk/error.go
+++ b/coder-sdk/error.go
@@ -17,7 +17,7 @@ var ErrPermissions = xerrors.New("insufficient permissions")
 // ErrAuthentication describes the error case in which the requester has invalid authentication.
 var ErrAuthentication = xerrors.New("invalid authentication")
 
-// APIError is the expected payload format for our errors.
+// APIError is the expected payload format for API errors.
 type APIError struct {
 	Err APIErrorMsg `json:"error"`
 }
--- a/coder-sdk/error.go
+++ b/coder-sdk/error.go
@@ -17,13 +17,13 @@ var ErrPermissions = xerrors.New("insufficient permissions")
 // ErrAuthentication describes the error case in which the requester has invalid authentication.
 var ErrAuthentication = xerrors.New("invalid authentication")
 
-// apiError is the expected payload format for our errors.
-type apiError struct {
-	Err apiErrorMsg `json:"error"`
+// APIError is the expected payload format for our errors.
+type APIError struct {
+	Err APIErrorMsg `json:"error"`
 }
 
-// apiErrorMsg contains the rich error information returned by API errors.
-type apiErrorMsg struct {
+// APIErrorMsg contains the rich error information returned by API errors.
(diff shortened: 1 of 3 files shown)


coder-sdk/error.go
```diff
@@ -17,31 +17,52 @@ var ErrPermissions = xerrors.New("insufficient permissions")
...
+	Msg     string          `json:"msg"`
+	Code    string          `json:"code"`
+	Details json.RawMessage `json:"details"`
 }
 
 // HTTPError represents an error from the Coder API.
 type HTTPError struct {
 	*http.Response
+	cached    *APIError
+	cachedErr error
```
Emyrk, line 36: We decode the `re.Response.Body` on `.Error()`. I didn't feel like trying to fix that behavior, but I need to cache the decoded payload as .Error() can be called more than once, and then the second one fails. So I split `.Error()` in to `.Payload` so I can extract the payload and not change any current behavior.

cmoog: oh nice catch, I also think we had this buggy code duplicated in the monorepo. merged a patch there last week.

coder-sdk/error.go
```diff
@@ -17,31 +17,52 @@ var ErrPermissions = xerrors.New("insufficient permissions")
 // ErrAuthentication describes the error case in which the requester has invalid authentication.
 var ErrAuthentication = xerrors.New("invalid authentication")
 
-// apiError is the expected payload format for our errors.
-type apiError struct {
-	Err apiErrorMsg `json:"error"`
+// APIError is the expected payload format for our errors.
+type APIError struct {
```
cmoog, line 21:
```suggestion
// APIError is the expected payload format for API errors.
type APIError struct {
```

cmoog: only being particular because making this public reveals its doc comment on godoc

cmoog (approved):

coder-sdk/error.go
Emyrk: It still feels a little odd, but I didn't want to refactor how errors worked as I only touched small areas of the cli so far.
So this works for now.

Emyrk:
I broke some test making the errors better... I'll look into it :/

Emyrk:
I realized this can be merged first, as I handle verbose errors better here too. So it works better against all release of `/m`

PR #272, opened 2021-03-09
fix: use proper API type for resource pool list

cmoog:


cmoog:
List view kube config looks like trash, will fix soon.

PR #275, opened 2021-03-09
Rename local pools to built-in

f0ssel:
for https://github.com/cdr/m/pull/7397


f0ssel:
@cmoog Is it OK if this integration test fails? Kinda a chicken and egg scenario between the PRs I think

cmoog:
Yeah that's fine, my bad. I've been meaning to parameterize the cemanager URL used by those tests and allow overriding in the workflow dispatch. but alas, one day.

cmoog (approved):

PR #277, opened 2021-03-10
Validate hostname, name, and cluster address on wsp create

f0ssel:
must be merged before https://github.com/cdr/m/pull/7417

### DIFF
--- a/internal/cmd/providers.go
+++ b/internal/cmd/providers.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"fmt"
 
+	"cdr.dev/coder-cli/internal/x/xcobra"
+
 	"github.com/spf13/cobra"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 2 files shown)


internal/cmd/providers.go
```diff
@@ -29,13 +28,18 @@ func providersCmd() *cobra.Command {
...
 
 func createProviderCmd() *cobra.Command {
+	var (
+		name           string
+		hostname       string
+		clusterAddress string
+	)
 	cmd := &cobra.Command{
-		Use:   "create [workspace_provider_name]",
+		Use:   "create --name=[name] --hostname=[hostname] --clusterAddress=[clusterAddress]",
```
cmoog, line 37: for consistency we may want to do `name` as a required positional argument, similar to `coder envs create`

deansheather (approved):

PR #278, opened 2021-03-10
chore: limit direct uses of os.Stdout/os.Stderr/os.Stdin

cmoog:
By moving the out/err/in channels to the cobra config,
it will be easier to add fast and isolated unit tests.


jawnsy (approved):
Sweet, LGTM :+1:

PR #279, opened 2021-03-10
chore: simple no auth unit test

cmoog:
More to come of course, but want to merge these many-file diffs in small PRs to reduce blocking and conflicts.

The next major inhibitor to test flexibility is the fact that our `config` values are mostly package-level. Will need to refactor.

### DIFF
--- a/pkg/tablewriter/tablewriter.go
+++ b/pkg/tablewriter/tablewriter.go
@@ -3,7 +3,6 @@ package tablewriter
 import (
 	"fmt"
 	"io"
-	"os"
 	"reflect"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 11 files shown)


pkg/tablewriter/tablewriter.go
```diff
@@ -55,19 +54,16 @@ func StructFieldNames(data interface{}) string {
 	return s.String()
 }
 
-// The output io.Writer for WriteTable. This is globally defined to allow overriding in tests.
```
jawnsy, line 58: Excellent fix :+1:

pkg/tablewriter/tablewriter_test.go
```diff
@@ -49,8 +49,7 @@ func TestTableWriter(t *testing.T) {
 	}
 
 	buf := bytes.NewBuffer(nil)
```
jawnsy, line 51: It doesn't matter at all, but the [bytes godoc](https://golang.org/pkg/bytes/#Buffer) seems to use `var buf bytes.Buffer` for some reason:

```go
package main

import (
	"bytes"
	"fmt"
	"os"
)

func main() {
	var b bytes.Buffer // A Buffer needs no initialization.
	b.Write([]byte("Hello "))
	fmt.Fprintf(&b, "world!")
	b.WriteTo(os.Stdout)
}
```

jawnsy (approved):
Always love PRs that add tests :heart_eyes:

pkg/tablewriter/tablewriter_test.go
cmoog: Will change this everywhere in the follow ups.

PR #280, opened 2021-03-10
chore: add auth init to unit tests

cmoog:


internal/cmd/cli_test.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	config.SetRoot(tmpDir)
+
+	// TODO: might need to make this a command scoped option to make assertions against its output
+	clog.SetOutput(ioutil.Discard)
+
+	email := os.Getenv("CODER_EMAIL")
+	password := os.Getenv("CODER_PASSWORD")
+	rawURL := os.Getenv("CODER_URL")
+	if email == "" || password == "" || rawURL == "" {
```
cmoog, line 37: @jawnsy strong feelings on this vs. flags?

jawnsy: I personally prefer flags, but whatever works I guess. The exception would be if these are standard variables that contain credentials (e.g. like KUBECONFIG) that we expect people to set, or that may be injected via the platform (e.g. if we inject tokens automatically into workspaces, so that coder-cli works out-of-the-box and is authenticated with the "owner" user credentials)

internal/cmd/cli_test.go
```diff
@@ -0,0 +1,135 @@
...
+	var cliErr clog.CLIError
+	if !xerrors.As(r.exitErr, &cliErr) {
+		slogtest.Fatal(t, "expected clog error, none found", slog.Error(r.exitErr), slog.F("type", fmt.Sprintf("%T", r.exitErr)))
+	}
+	slogtest.Debug(t, "clog error", slog.F("message", cliErr.String()))
+	return cliErr
+}
+
+func execute(t *testing.T, in io.Reader, args ...string) result {
+	cmd := Make()
```
jawnsy, line 111: should we use t.Helper() for these? I guess it doesn't matter as we don't use assertions in here, it may be useful for the other functions though

internal/cmd/envs_test.go
```diff
@@ -1,76 +1,18 @@
...
-type result struct {
-	OutBuffer *bytes.Buffer
-	ErrBuffer *bytes.Buffer
-	ExitErr   error
-}
+func Test_envs_ls(t *testing.T) {
+	res := execute(t, nil, "envs", "ls")
+	res.success(t)
 
-func execute(t *testing.T, args []string, in io.Reader) result {
```
jawnsy, line 42: Nice cleanup :+1:

jawnsy (approved):

internal/cmd/cli_test.go
cmoog: Oh true true, added those where appropriate.

PR #282, opened 2021-03-11
feature: support filtering by workspace provider when listing environ…

fuskovic:
# What this does?

Adds the `--provider` flag to both the `coder envs ls` command as well as the `coder resources top` command.

## envs command

Now supports the `--provider` flag so users can filter environments by a particular workspace provider name.

Screenshot 2021-03-11 at 3 22 46 PM

## resource manager command

Now supports the `--provider` flag so users can aggregate resource amounts by a particular workspace provider name.

Screenshot 2021-03-11 at 3 52 41 PM

### DIFF
--- a/ci/integration/envs_test.go
+++ b/ci/integration/envs_test.go
@@ -144,6 +144,20 @@ func TestEnvsCLI(t *testing.T) {
 			tcli.StdoutMatches(regexp.QuoteMeta(name)),
 		)
 
+		// filter by provider that does not exist should fail
+		doesntExist := randString(10)
+		c.Run(ctx, fmt.Sprintf("coder envs ls --provider %s", doesntExist)).Assert(t,
+			tcli.Error(),
+			tcli.StderrMatches(regexp.QuoteMeta(fmt.Sprintf("fatal: no environments found for workspace provider %q", doesntExist))),
+		)
+
+		// filter by provider that does exist should succeed
+		var envs []coder.Environment
+		c.Run(ctx, "coder envs ls --provider built-in").Assert(t,
+			tcli.Success(),
+			tcli.StdoutJSONUnmarshal(&envs),
+		)
+
 		var env coder.Environment
 		c.Run(ctx, fmt.Sprintf(`coder envs ls -o json | jq '.[] | select(.name == "%s")'`, name)).Assert(t,
 			tcli.Success(),
--- a/ci/integration/envs_test.go
+++ b/ci/integration/envs_test.go
@@ -144,20 +144,6 @@ func TestEnvsCLI(t *testing.T) {
 			tcli.StdoutMatches(regexp.QuoteMeta(name)),
 		)
 
-		// filter by provider that does not exist should fail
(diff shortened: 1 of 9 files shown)


ci/integration/envs_test.go
```diff
@@ -144,6 +144,20 @@ func TestEnvsCLI(t *testing.T) {
...
+			tcli.StderrMatches(regexp.QuoteMeta(fmt.Sprintf("fatal: no environments found for workspace provider %q", doesntExist))),
+		)
+
+		// filter by provider that does exist should succeed
+		var envs []coder.Environment
+		c.Run(ctx, "coder envs ls --provider built-in").Assert(t,
+			tcli.Success(),
+			tcli.StdoutJSONUnmarshal(&envs),
+		)
+
```
cmoog, line 160: as sad as it makes me I think we'd be best served to move these from `tcli` into `internal/cmd/envs_test.go` with the new in-memory framework.

internal/cmd/envs.go
```diff
@@ -51,6 +51,7 @@ func lsEnvsCommand() *cobra.Command {
 	var (
 		outputFmt string
 		user      string
+		provider  string
 	)
 
 	cmd := &cobra.Command{
```
cmoog, line 57: it would also be very nice if we included `Provider` as a table column in all calls to `envs ls`... I can follow this up with that change though.

internal/cmd/ceapi.go
```diff
@@ -202,3 +202,30 @@ func getUserOrgs(ctx context.Context, client coder.Client, email string) ([]code
 	}
 	return lookupUserOrgs(u, orgs), nil
 }
+
+func getProviderByName(ctx context.Context, client coder.Client, wpName string) (*coder.KubernetesProvider, error) {
```
cmoog, line 206: looks like we already have this in `coderutil.ProviderByName`

cmoog:
So I'm having second thoughts about this design. In all other cases `coder envs ls` only lists environments _owned_ by the active user. I think this functionality should be rolled under the `coder resources` subcommand where the use case more closely matches.

internal/cmd/resourcemanager.go
```diff
@@ -287,6 +319,14 @@ func (u userLabeler) label(e coder.Environment) string {
 	return fmt.Sprintf("[user: %s]", u.userMap[e.UserID].Email)
 }
 
+type providerLabeler struct {
+	providerMap map[string]coder.KubernetesProvider
+}
+
+func (p providerLabeler) label(e coder.Environment) string {
+	return fmt.Sprintf("[provider %s]", p.providerMap[e.ResourcePoolID].Name)
+}
```
cmoog, line 328:
```suggestion
func (p providerLabeler) label(e coder.Environment) string {
	return fmt.Sprintf("[provider: %s]", p.providerMap[e.ResourcePoolID].Name)
}
```

cmoog (approved):
great stuff. looks good to go

PR #285, opened 2021-03-11
Update create-from-config for compatibility with remote templates

sreya:


coder-sdk/interface.go
```diff
@@ -79,10 +79,10 @@ type Client interface {
 
 	// ParseTemplate parses a template config. It support both remote repositories and local files.
 	// If a local file is specified then all other values in the request are ignored.
-	ParseTemplate(ctx context.Context, req ParseTemplateRequest) (Template, error)
+	ParseTemplate(ctx context.Context, req ParseTemplateRequest) (TemplateVersion, error)
```
cmoog, line 82: This should probably return `*TemplateVersion` to match the rest of `coder-sdk`

cmoog (approved):
Looks good

PR #292, opened 2021-03-17
feat: replace `coder sh` implementation to shell out to `ssh`

cmoog:
Previously, we used our custom WebSocket shell protocol
to open shells with `coder sh`. This divergent behavior
from genuine SSH created confusion and a lack of feature parity
between the two connection methods.


internal/cmd/cmd.go
```diff
@@ -24,7 +24,7 @@ func Make() *cobra.Command {
 	app.AddCommand(
 		loginCmd(),
 		logoutCmd(),
-		shCmd(),
```
deansheather, line 27: We cannot remove this command because SSH is optional.

deansheather (dismissed):

jawnsy:
Is this going to affect dependencies required on Windows? I think Linux and Mac machines often include OpenSSH clients, but the same is not true on Windows. I'm on my phone right now but will take a deeper look later

internal/cmd/cmd.go
cmoog: I'm working on some docs and supplementary material, but I basically think that is a bad requirement and doesn't make sense anymore.

deansheather:
@jawnsy windows 10 has openssh installed by default now

internal/cmd/cmd.go
deansheather: SSH being optional will probably need to be deprecated and not removed outright since it's a major change. Keeping the old shell command will be necessary until SSH is required.

cmoog:
Obviously as follow-ups we could add flag behavior similar to `gcloud compute ssh`...
```
ERROR: (gcloud.compute.ssh) argument [USER@]INSTANCE: Must be specified.
Usage: gcloud compute ssh [USER@]INSTANCE [optional flags] [-- SSH_ARGS ...]
  optional flags may be  --command | --container | --dry-run |
                         --force-key-file-overwrite | --help | --internal-ip |
                         --plain | --ssh-flag | --ssh-key-expiration |
                         --ssh-key-expire-after | --ssh-key-file |
                         --strict-host-key-checking | --tunnel-through-iap |
                         --zone

For detailed information on this command and its flags, run:
  gcloud compute ssh --help
```

internal/cmd/ssh.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+func sshCmd() *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "ssh [environment_name] [<command [args...]>]",
+		Short: "Enter a shell of execute a command over SSH into a Coder environment",
+		Args:  shValidArgs,
+		Example: `coder ssh my-dev
+coder ssh my-dev pwd`,
+		Aliases:               []string{"sh"},
```
sreya, line 30: 👍

sreya (approved):

PR #297, opened 2021-03-22
chore: migrate more tests from tcli to in-mem

cmoog:


internal/cmd/envs_test.go
```diff
@@ -45,3 +56,151 @@ func Test_envs_ls_by_provider(t *testing.T) {
...
+}
+
+var floatComparer = cmp.Comparer(func(x, y float64) bool {
+	delta := math.Abs(x - y)
+	mean := math.Abs(x+y) / 2.0
+	return delta/mean < 0.001
+})
+
+// this is a stopgap until we have support for a `coder images` subcommand
+// until then, we want can use the coder.Client to ensure our integration tests
```
fuskovic, line 153:
```suggestion
// until then, we can use the coder.Client to ensure our integration tests
```

internal/cmd/envs_test.go
```diff
@@ -45,3 +56,151 @@ func Test_envs_ls_by_provider(t *testing.T) {
...
+			if m.Email == os.Getenv("CODER_EMAIL") {
+				o := o
+				org = &o
+				break search
+			}
+		}
+	}
+	if org == nil {
+		slogtest.Fatal(t, "failed to find org of current user")
+		return // help the linter out a bit
```
fuskovic, line 172: Seems strange to have a return after a fatal but I guess there's something going on with the linter that's not obvious to me. The comment doesn't  help explain why so I think we can just elaborate on the comment or just add a `nolint`.  Either way no big deal just my general thoughts reading.

internal/cmd/cli_test.go
```diff
@@ -146,6 +146,10 @@ func execute(t *testing.T, in io.Reader, args ...string) result {
 	cmd.SetOut(&outStream)
 	cmd.SetErr(&errStream)
 
+	// TODO: this *needs* to be moved to function scoped writer arg. As is,
+	// this prevents tests from running in parallel.
```
fuskovic, line 150: good to know

fuskovic (approved):

PR #299, opened 2021-03-23
fix: properly label by non grouped property

cmoog:
So the difference here is rather subtle and bit counterintuitive. We want to "label" by the property which is not being grouped by. So in this case we only want to "label" the provider when we are grouping my user/org. 

It's completely redundant to label each environment with the provider when they are already grouped together and have the provider in the group header.


cmoog:
This can be followed-up with the addition of the `image` and `tag` labels...

PR #302, opened 2021-03-24
feat: implement cordon/uncordon workspace provider

fuskovic:
# What this does?

Implements two new subcommands.

### Usage

`coder providers cordon --reason "cloud cost" provider-name`

Allows users to cordon a provider; preventing it from supporting new workspaces.

**Note** : the `--reason` flag is required to explain why this action was performed.

### Usage

`coder providers uncordon provider-name`

Also supports uncordoning a provider; allowing the provider to continue provisioning resources for new workspaces.

## Sample Output

Screenshot 2021-03-24 at 8 05 58 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for Coder CLI 
 
-.PHONY: clean build build/macos build/windows build/linux fmt lint gendocs test/go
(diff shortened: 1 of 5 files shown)


f0ssel (approved):
looks great

PR #304, opened 2021-03-30
chore: rename instances of Coder Enterprise to Coder

jawnsy:
We rebranded the product from Coder Enterprise to Coder; this change
updates the text accordingly. Some references (such as URLs and
variable names) have not been changed.


internal/cmd/configssh.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const sshStartToken = "# ------------START-CODER-ENTERPRISE-----------"
```
jawnsy, line 23: We can change these, but some of the code looks for this token, so we'll have to handle both cases (old heading and new heading) and I'm not sure if it's worth it

PR #305, opened 2021-03-30
Add confirmation prompt to 'coder envs edit'

lilshoff:
See CH9720 https://app.clubhouse.io/coder/story/9720/confirm-when-editing-environment-with-coder-cli

### DIFF
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -485,7 +485,7 @@ coder envs edit back-end-env --disk 20`,
 
 			if !force && env.LatestStat.ContainerStatus == coder.EnvironmentOn {
 				_, err = (&promptui.Prompt{
-					Label:     fmt.Sprintf("Rebuild environment \"%s\"? (will destroy any work outside of your home directory)", env.Name),
+					Label:     fmt.Sprintf("Rebuild environment %q? (will destroy any work outside of your home directory)", env.Name),
 					IsConfirm: true,
 				}).Run()
 				if err != nil {
--- a/internal/cmd/envs.go
+++ b/internal/cmd/envs.go
@@ -430,6 +430,7 @@ func editEnvCmd() *cobra.Command {
 		gpus   int
 		follow bool
 		user   string
+		force  bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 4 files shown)


internal/cmd/envs.go
```diff
@@ -430,6 +430,7 @@ func editEnvCmd() *cobra.Command {
 		gpus   int
 		follow bool
 		user   string
+		force  bool
```
jawnsy, line 433: nice :+1:

internal/cmd/envs.go
```diff
@@ -482,6 +483,19 @@ coder envs edit back-end-env --disk 20`,
 				return err
 			}
 
+			if !force && env.LatestStat.ContainerStatus == coder.EnvironmentOn {
+				_, err = (&promptui.Prompt{
+					Label:     fmt.Sprintf("Rebuild environment \"%s\"? (will destroy any work outside of /home)", env.Name),
```
jawnsy, line 488: it doesn't matter for anything, but fyi you can use `%q` to double quote your string (you can tell that I am extremely lazy)

```suggestion
					Label:     fmt.Sprintf("Rebuild environment %q? (will destroy any work outside of /home)", env.Name),
```

You might also want to mention "your home directory" instead of a specific path, because as a special case, root's home directory is `/root`

internal/cmd/envs_test.go
```diff
@@ -108,7 +108,7 @@ func Test_env_create(t *testing.T) {
 
 	// edit the CPU of the environment
 	cpu = 2.1
-	res = execute(t, nil, "envs", "edit", name, fmt.Sprintf("--cpu=%f", cpu), "--follow")
+	res = execute(t, nil, "envs", "edit", name, fmt.Sprintf("--cpu=%f", cpu), "--follow", "--force")
```
jawnsy, line 111: love that you added a test, TY :heart:

jawnsy (approved):
LGTM! :fire:

internal/cmd/envs.go
vapurrmaid: I agree on "your home directory"

vapurrmaid (approved):

internal/cmd/envs.go
lilshoff: Oh nice - didn't know about that `%q` 🤯 🙌 

And yeah the rename from `/home` to `your home directory` is a good call. Fixed it on the rebuild prompt too 👌

PR #308, opened 2021-03-31
Consider template type on 'envs edit'

lilshoff:
CLI updates to environments created via remote/local templates should not succeed.
Currently, the cli _silently_ fails to update these environments. It should display an error and tell the user how to correctly update these types of environments. 

This PR checks the template type of the environment before attempting to update, and exits with helpful tip if the environment was not created via the legacy web ui.


internal/cmd/envs.go
```diff
@@ -457,6 +458,22 @@ coder envs edit back-end-env --disk 20`,
 				return err
 			}
 
+			// CLI updates for environments that were created via templates should not work
+			tt, err := client.QueryTemplate(ctx, env.TemplateVersionID)
```
Emyrk, line 462: Doesn't the `templateVersion` return the version being used and not a list?

lilshoff: Not really sure what you mean by that tbh

---

But taking a second look at the code in `m`
This probably shouldn't return a list: https://github.com/cdr/m/blob/master/product/coder/pkg/model/template.go#L723

But yeah the handler for QueryTemplate returns a list (and probably shouldn't)
https://github.com/cdr/m/blob/c3e5451cb5457641aaa6f5858d4c1ea2bb1ae7b0/product/coder/service/cemanager/templates.go#L389

Is that what you were getting at?

sreya: The intent of `QueryEndpoints` is to be a generic endpoint for getting templates based on query criteria. I had it return a list so that when we want to be able to query templates by, for example orgs, we don't need to have a separate endpoint. Querying by templateVersion should obviously return a list of size 1.

Emyrk: Yea because we query by template version, I would expect a list of length 1

lilshoff:
Yo we good w this one? 
@Emyrk @sreya

Emyrk (approved):
I still think that last code should always just return 1 template, but the sort and all to make sure is fine

PR #311, opened 2021-04-06
Add coder agent start

f0ssel:
Add a command to allow users to start the environment p2p agent. This command is hidden for now while we experiment.

```
# Uses CODER_AGENT_TOKEN implicitly
coder agent start https://master.cdr.dev
# Or pass as flag
coder agent start https://master.cdr.dev --token xxx-xxx
```


internal/cmd/agent.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/pion/webrtc/v3"
+	"github.com/spf13/cobra"
+	"golang.org/x/xerrors"
+	"io"
+	"net"
+	"net/url"
+	"nhooyr.io/websocket"
+	"os"
+	"strings"
+	"time"
```
deansheather, line 22: please sort these into 3 groups

internal/cmd/agent.go
```diff
@@ -0,0 +1,259 @@
...
+			if token == "" {
+				var ok bool
+				token, ok = os.LookupEnv("CODER_AGENT_TOKEN")
+				if !ok {
+					return xerrors.New("must pass --token or set the CODER_AGENT_TOKEN env variable")
+				}
+			}
+
+			q := u.Query()
+			q.Set("agent_token", token)
```
deansheather, line 81:
```suggestion
			q.Set("service_token", token)
```

internal/cmd/agent.go
```diff
@@ -0,0 +1,259 @@
...
+
+			ctx, cancelFunc := context.WithTimeout(context.Background(), time.Second*15)
+			defer cancelFunc()
+			log.Info(ctx, "connecting to broker", slog.F("url", u.String()))
+			conn, _, err := websocket.Dial(ctx, u.String(), nil)
+			if err != nil {
+				return fmt.Errorf("dial: %w", err)
+			}
+			nc := websocket.NetConn(context.Background(), conn, websocket.MessageBinary)
+			session, err := yamux.Server(nc, nil)
```
deansheather, line 92: nitpicky since it doesn't really matter since yamux is bidirectional, but this should be a client because it's on the client end of a websocket connection

internal/cmd/agent.go
```diff
@@ -0,0 +1,259 @@
...
+			u.RawQuery = q.Encode()
+
+			ctx, cancelFunc := context.WithTimeout(context.Background(), time.Second*15)
+			defer cancelFunc()
+			log.Info(ctx, "connecting to broker", slog.F("url", u.String()))
+			conn, _, err := websocket.Dial(ctx, u.String(), nil)
+			if err != nil {
+				return fmt.Errorf("dial: %w", err)
+			}
+			nc := websocket.NetConn(context.Background(), conn, websocket.MessageBinary)
```
deansheather, line 91: I know this is an MVP but there's a lot going on in this command so all of this should move into it's own package (or potentially even it's own repo if we want to use it within our monorepo as well).

deansheather (commented):
I don't know much about RTC stuff so I'm not the best person to review that portion of this PR

internal/cmd/agent.go
f0ssel: @kylecarbs Any objections? (since this part is your code)

internal/cmd/agent.go
f0ssel: I pinky promise I'll do that if this ends up being a thing we stick with ;)

nathanpotter:
this is incredible 🎉

internal/cmd/agent.go
```diff
@@ -0,0 +1,270 @@
...
+	)
+	return cmd
+}
+
+func startCmd() *cobra.Command {
+	var (
+		token string
+	)
+	cmd := &cobra.Command{
+		Use:   "start [coderURL] --token=[token]",
```
nathanpotter, line 46: do you think `coderURL` should be a `--url` flag? I'm imagining a case where this is defaulted to `coder.com` in the future for a SaaS offering. However, I know this isn't ideal for now since the `coder agent start` cmd would fail without the flags specified, so more so opening this up for discussion. Could potentially default `coderURL` to `coder.com` in the future as well if it's not specified as another option.

f0ssel: Yeah this is _suuuper_ mvp just to get a POC working in coder, a real implementation will follow once we get an idea of how scalable and stable this is.

f0ssel:
I can't fix the `fmt` step because when I run `make -j fmt` on my env there are no changes. I guess @deansheather would you know what I can do?

deansheather (approved):

PR #316, opened 2021-04-15
Add retries to coder agent start cmd

f0ssel:
This organizes the code to into a new `agent` pkg, in hopes we can share some code with the coder repo as well when the time comes.

I wrap the main server in a retry loop to make sure the agent is always running (if it get's passed the input parsing in the command)

### DIFF
--- a/internal/cmd/agent.go
+++ b/internal/cmd/agent.go
@@ -2,25 +2,15 @@ package cmd
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
-	"io"
-	"net"
 	"net/url"
 	"os"
-	"strings"
-	"time"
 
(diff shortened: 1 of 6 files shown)


internal/cmd/agent.go
```diff
@@ -60,7 +50,7 @@ coder agent start --coder-url https://my-coder.com --token xxxx-xxxx
 
 			if coderURL == "" {
 				var ok bool
-				token, ok = os.LookupEnv("CODER_URL")
+				coderURL, ok = os.LookupEnv("CODER_URL")
```
f0ssel, line 53: This was a bug

deansheather (approved):

kylecarbs (approved):

PR #322, opened 2021-04-22
feat: Use internal TURN server for proxying

kylecarbs:
This feature is under a feature flag in the product at a deployment level, so there's no risk of security gaps here.

### DIFF
--- a/internal/cmd/tunnel.go
+++ b/internal/cmd/tunnel.go
@@ -110,7 +110,11 @@ type client struct {
 
 func (c *client) start() error {
 	url := fmt.Sprintf("%s%s%s%s%s", c.brokerAddr.String(), "/api/private/envagent/", c.id, "/connect?session_token=", c.token)
-	tcpProxy := fmt.Sprintf("turns:%s:5349?transport=tcp", c.brokerAddr.Host)
+	turnScheme := "turns"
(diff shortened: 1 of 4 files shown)


internal/cmd/tunnel.go
```diff
@@ -108,9 +109,9 @@ type client struct {
 }
 
 func (c *client) start() error {
-	url := fmt.Sprintf("%s%s%s%s%s", c.brokerAddr, "/api/private/envagent/", c.id, "/connect?session_token=", c.token)
-	c.logger.Info(c.ctx, "connecting to broker", slog.F("url", url))
-
+	url := fmt.Sprintf("%s%s%s%s%s", c.brokerAddr.String(), "/api/private/envagent/", c.id, "/connect?session_token=", c.token)
+	tcpProxy := fmt.Sprintf("turns:%s:5349?transport=tcp", c.brokerAddr.Host)
```
f0ssel, line 113: `turns` is using tls right? We could technically look at the broker addr scheme here and decide if it's TURN or TURNs.

f0ssel (approved):

internal/cmd/tunnel.go
kylecarbs: Good point... I'll change this now!

PR #326, opened 2021-04-27
Centralize p2p Dial logic

f0ssel:
This creates a WorkspaceClient that can dial workspaces and represent data channels as net.Conns. With this abstraction we can dial wsep commands over these connections, which is my next step.


xwebrtc/workspace.go
```diff
@@ -0,0 +1,236 @@
...
+// DialContext will create a new peer connection with the workspace agent, make a new data channel, and return it as
+// a net.Conn.
+func (wc *WorkspaceClient) DialContext(ctx context.Context, network string, workspaceAddr string) (net.Conn, error) {
+	workspaceID, port, err := ParseWorkspaceAddr(workspaceAddr)
+	if err != nil {
+		return nil, xerrors.Errorf("parsing workspace addr: %w", err)
+	}
+
+	wc.log.Debug(ctx, "parsed wsp addr", slog.F("id", workspaceID), slog.F("port", port))
+	rtc, err := wc.PeerConnection(ctx, workspaceID)
```
kylecarbs, line 65: I'd rather if we made the `WorkspaceClient` create the connection on `NewWorkspaceClient` for a specific workspace, then our `DialContext` arguments can be normal instead of a customized format.

kylecarbs (changes requested):

xwebrtc/workspace.go
f0ssel: Updated to make the active peer connection in the workspace client and renamed to WorkspaceDialer

xwebrtc/client.go
```diff
@@ -0,0 +1,32 @@
...
+	token          string
+}
+
+func NewWorkspaceClient(log slog.Logger, brokerAddr *url.URL, token string) *WorkspaceClient {
+	return &WorkspaceClient{
+		log: log,
+		brokerAddr: brokerAddr,
+		token: token,
+	}
+}
```
kylecarbs, line 23: Does this need to exist anymore? We could just add these args to `DialWorkspace` imo

xwebrtc/channel.go
```diff
@@ -0,0 +1,75 @@
...
+	"golang.org/x/xerrors"
+	"strings"
+	"time"
+
+	"github.com/pion/webrtc/v3"
+)
+
+// WaitForDataChannelOpen waits for the data channel to have the open state.
+// By default, it waits 15 seconds.
+func WaitForDataChannelOpen(ctx context.Context, channel *webrtc.DataChannel) error {
```
kylecarbs, line 15: I don't think we should expose any funcs we don't need to here for rn

xwebrtc/client.go
f0ssel: gone

kylecarbs (approved):
Bomb API

PR #329, opened 2021-05-02
feat: Durability test RTC connections

kylecarbs:
Hardens and fully E2E tests brokering of connections.

We can use the `DialICE` function to ensure the TURN server is configured properly during setup too...

New API example here:
https://github.com/cdr/coder-cli/blob/1d7d14c51ff0af06ddaf8f18e013dee9ee4bc1b9/wsnet/dial_test.go#L12-L41

[ch11739]


f0ssel (approved):
this all looks amazing

PR #338, opened 2021-05-09
fix: Stop infinite read loop on control channel

kylecarbs:
Big yikers!


kylecarbs:
Merging without approval... I've tested this to ensure it works. Pretty necessary for not using all the CPU!

jawnsy (approved):

jawnsy:
@kylecarbs I don't think we should merge without reviews, except in emergencies; it's not a good habit for us to get into.

PR #344, opened 2021-05-11
chore: update golangci-lint configuration

jawnsy:


f0ssel (approved):
not sure what all the diff really is but I trust

jawnsy:
@f0ssel yea it was a lot of changes, I tried to format things to better match the monorepo settings, and also enabled more linters as long as they were passing. I removed exclusions which don't seem to be needed anymore (I guess we fixed code in the meantime)

PR #349, opened 2021-05-13
fix: Stop looping retries after dial failure

kylecarbs:
Pretty unfortunate bug here... noticed while testing.

After the initial dial the loop was never exited, causing infinite redials.

### DIFF
--- a/wsnet/listen.go
+++ b/wsnet/listen.go
@@ -49,6 +49,9 @@ func Listen(ctx context.Context, broker string) (io.Closer, error) {
 					case <-ctx.Done():
 						err = ctx.Err()
 					}
+					if err == nil {
+						break
+					}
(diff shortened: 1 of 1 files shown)


wsnet/listen.go
```diff
@@ -49,6 +49,9 @@ func Listen(ctx context.Context, broker string) (io.Closer, error) {
 					case <-ctx.Done():
 						err = ctx.Err()
 					}
+					if err == nil {
+						break
+					}
 					if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 						break
 					}
```
deansheather, line 57:
```suggestion
					if err == nil || errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
						break
					}
```

deansheather (approved):

PR #350, opened 2021-05-13
Rename environments to workspaces and use new routes

code-asher:
Also closes [ch5518].


f0ssel (approved):
Has anyone else been looped in on the command change from `coder envs` to `coder ws`? I know we have been using that shorthand internally but since this is an customer command I'd wanna make sure others are on board. I don't hate `workspaces` but I do see it's long.

kylecarbs:
We should do both workspaces and ws imo.

code-asher:
Currently I have `workspaces` as an alias for `ws` but I think the reverse might actually make more sense.

code-asher:
Swapped! We use `workspaces` everywhere now including documentation but `ws` is a valid alias to use in its place. I think this is quite a bit more clear especially since `ws` often means `websocket`, at least in the codebase.

kylecarbs:
✨

f0ssel (approved):
:pray:

PR #352, opened 2021-05-17
fix: Drop port when creating TURN endpoint

kylecarbs:
Caused strings like `localhost:5349:8080` when a hostname contained a port.


f0ssel (approved):
TIL there is a difference

PR #353, opened 2021-05-19
coder: remove some unnecessary SSH settings

sreya:
- Removed ServerAliveInterval and ServerAliveCountMax from config-ssh
  settings. These settings are opinionated about when idle sessions should disconnect
  and should be inherited from the user's global settings instead. 'coder
  config-ssh' should only write configuration options that are necessary
  for 'coder sh' and its variants to succeed.


kylecarbs (approved):
Seems fine to me.

jawnsy:
[ch12778]

PR #357, opened 2021-05-21
Warn if coder-cli is not in PATH on Windows

deansheather:
In Powershell and VS Code, the OpenSSH for Windows `ProxyCommand` is expected to be a binary name like `coder` or a Windows-like path such as `C:\path\to\coder.exe`.

In Msys2 and Git Bash, the Linux-y OpenSSH expects the `ProxyCommand` to be a binary name or a Unix-like path such as `/c/path/to/coder`.

To support both environments simultaneously, the `ProxyCommand` must be set to `coder` (or whatever binary name coder-cli has), and we must warn the user if the binary is not in `$PATH`.

[CH-15394]

### DIFF
--- a/ci/scripts/build.sh
+++ b/ci/scripts/build.sh
@@ -8,11 +8,17 @@ set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)/ci/scripts"
 
-tag=$(git describe --tags)
+tag="$(git describe --tags)"
 
-echo "--- building coder-cli for $GOOS-$GOARCH"
+flavor="$GOOS"
+if [[ "$GOOS" == "windows" ]]; then
+	unset GOARCH
+else
+	flavor+="-$GOARCH"
+fi
+echo "--- building coder-cli for $flavor"
 
-tmpdir=$(mktemp -d)
+tmpdir="$(mktemp -d)"
 go build -ldflags "-X cdr.dev/coder-cli/internal/version.Version=${tag}" -o "$tmpdir/coder" ../../cmd/coder
 
 cp ../gon.json $tmpdir/gon.json
@@ -20,7 +26,7 @@ cp ../gon.json $tmpdir/gon.json
 pushd "$tmpdir"
 case "$GOOS" in
 "windows")
-	artifact="coder-cli-$GOOS-$GOARCH.zip"
+	artifact="coder-cli-$GOOS.zip"
 	mv coder coder.exe
 	zip "$artifact" coder.exe
 	;;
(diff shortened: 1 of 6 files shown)


internal/cmd/configssh.go
```diff
@@ -165,6 +166,52 @@ func configSSH(configpath *string, remove *bool, next *bool) func(cmd *cobra.Com
...
+	// On Windows, the coder-cli executable must be in $PATH for both Msys2/Git
+	// Bash and OpenSSH for Windows (used by Powershell and VS Code) to function
+	// correctly. Check if the current executable is in $PATH, and warn the user
+	// if it isn't.
+	if runtime.GOOS == "windows" {
+		binName := filepath.Base(exePath)
+		// We use safeexec instead of os/exec because os/exec returns paths in
+		// the current working directory, which we will run into very often when
+		// looking for our own path.
+		pathPath, err := safeexec.LookPath(binName)
```
deansheather, line 186: safeexec works on my windows machine in a test program I wrote, but never seems to work in the coder-cli cross-compiled builds from my workspace. I wonder if it's caused by cross-compile or if there's some BS at play

kylecarbs:
What's blocking this? Seems like it could be important for the `coder tunnel`

deansheather:
@kylecarbs The package I'm using to check if the binary is in path seems to fail when cross-compiled for Windows from within my workspace, but works when I build on my Windows machine. I may just write my own function that does the same thing.

deansheather:
It works when I remove `GOARCH=386` from the windows builds.

ci/scripts/build.sh
```diff
@@ -8,19 +8,25 @@ set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)/ci/scripts"
 
-tag=$(git describe --tags)
+tag="$(git describe --tags)"
 
-echo "--- building coder-cli for $GOOS-$GOARCH"
+flavor="$GOOS"
+if [[ "$GOOS" == "windows" ]]; then
+	unset GOARCH
```
kylecarbs, line 15: Probably worth adding a comment here.

internal/cmd/configssh.go
```diff
@@ -212,7 +259,7 @@ func makeSSHConfig(binPath, host, userName, workspaceName, privateKeyFilepath st
 		host := fmt.Sprintf(
 			`Host coder.%s
    HostName coder.%s
-   ProxyCommand %s tunnel %s 12213 stdio
+   ProxyCommand "%s" tunnel %s 12213 stdio
```
kylecarbs, line 262: Do the quotes work on Linux too? I _thought_ I tried it before...

kylecarbs (approved):

internal/cmd/configssh.go
deansheather: Works on windows, mac and linux from my testing.

PR #360, opened 2021-05-25
NetV2: provider create and install

f0ssel:
`coder providers install` - Will install the service accounts, role, and role bindings for a workspace provider in the current kube context

Example:
```
➜  coder-cli git:(wsp-install) ✗ go run cmd/coder/main.go providers install

Current Kubernetes context details:

Context Name: gke_coder-dev-1_us-central1-a_dev-2
Namespace:    f0ssel-install

The following kubernetes resources will be created or updated:

-- ServiceAccount
Name: coder
Namespace: f0ssel-install

-- ServiceAccount
Name: coder-workspaces
Namespace: f0ssel-install

-- Role
Name: coder
Namespace: f0ssel-install
Rules: 
  - APIGroups: "",networking.k8s.io
    Resources: persistentvolumeclaims,pods,secrets,pods/exec,pods/log,events,networkpolicies
    Verbs: create,get,list,watch,update,patch,delete,deletecollection
  - APIGroups: metrics.k8s.io,storage.k8s.io
    Resources: pods
    Verbs: get,list,watch

-- RoleBinding
Role:
  Kind:  Role
  Name:  coder
Subjects:
  - Kind: ServiceAccount
    Name: coder
    Namespace: f0ssel-install

Press the Enter Key to continue

Installing kubernetes resources...
Successfully installed kubernetes resources
```


go.mod
```diff
@@ -26,6 +26,8 @@ require (
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d
 	golang.org/x/time v0.0.0-20210220033141-f8bda1e9f3ba
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
+	k8s.io/api v0.21.1
+	k8s.io/apimachinery v0.21.1
 	k8s.io/client-go v0.21.1
```
jawnsy, line 31: hmm, I wonder why the CLI needs k8s APIs? very strange to me

f0ssel: I've updated the description and code to give more context

jawnsy: Thanks! One thing I'm not sure about is how close the versions of these libraries need to be to the corresponding Kubernetes versions, because we need to make sure that the versions match the Kubernetes versions that we support.

So, will v0.21.1 - which I guess corresponds to Kubernetes 1.21.1 - work with a Kubernetes 1.13 cluster?

f0ssel: no clue, but I can just lock it with the same version as the coder app

internal/x/xkube/provider.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// CurrentKubeContext returns the details of the current kubernetes context.
+func CurrentKubeContext() (*KubeContext, error) {
+	clientCfg, err := clientcmd.NewDefaultClientConfigLoadingRules().Load()
+	if err != nil {
+		return nil, xerrors.Errorf("loading kubernetes client config: %w", err)
+	}
+	currentContext := clientCfg.Contexts[clientCfg.CurrentContext]
+	cluster := clientCfg.Clusters[currentContext.Cluster]
```
deansheather, line 44: should do `, ok :=` for these

internal/x/xkube/provider.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// CoderServiceAccountFromContext reads the current kube context and fetches the coder workspace provider service account.
+func CoderServiceAccountFromContext(ctx context.Context) (*ServiceAccount, error) {
+	kctx, err := CurrentKubeContext()
+	if err != nil {
+		return nil, xerrors.Errorf("getting current kube context: %w", err)
+	}
+
+	saSecrets, err := kctx.Clientset.CoreV1().Secrets(kctx.Namespace).List(ctx, metav1.ListOptions{})
```
deansheather, line 77: Should add a comment here that we can't use an annotation field selector because it doesn't exist. I checked corev1 API to see if this could be done via ListOptions but it can't

internal/x/xkube/provider.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, rule := range roleSpec().Rules {
+		// core api group is represented as "" so we handle that formatting
+		for i, group := range rule.APIGroups {
+			if group == "" {
+				rule.APIGroups[i] = `""`
+			}
+		}
+		groups := strings.Join(rule.APIGroups, ",")
+		resources := strings.Join(rule.Resources, ",")
+		verbs := strings.Join(rule.Verbs, ",")
```
deansheather, line 131:
```suggestion
		groups := strings.Join(rule.APIGroups, ", ")
		resources := strings.Join(rule.Resources, ", ")
		verbs := strings.Join(rule.Verbs, ", ")
```

internal/x/xkube/provider.go
```diff
@@ -0,0 +1,251 @@
...
+			if err != nil {
+				return xerrors.Errorf("updating service account: %w", err)
+			}
+			return nil
+		}
+		return xerrors.Errorf("creating service account: %w", err)
+	}
+
+	return nil
+}
```
deansheather, line 169: could dedup these by making a helper function since they're mostly identical other than which spec it'll use

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
-				return xerrors.Errorf("create workspace provider: %w", err)
+				return xerrors.Errorf("reading kubernetes context: %w", err)
+			}
+			fmt.Printf(`
+Current Kubernetes context details:
+
+Context Name: %s
+Namespace:    %s
+
+The following kubernetes resources will be created or updated:
```
deansheather, line 152:
```suggestion
The following kubernetes resources will be created or replaced:
```

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
-				ClusterAddress: clusterAddress,
-			}
+func installProviderCmd() *cobra.Command {
+	var (
+		skipPrompt bool
+	)
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "Install kubernetes resources for a new workspace provider.",
+		Long:  "Install kubernetes resources for a new workspace provider.",
```
deansheather, line 132:
```suggestion
		Long:  `Install kubernetes resources for a new workspace provider.

This uses your current kubectl context to install the necessary resources for a Kubernetes workspace provider onto the cluster.`,
```

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
+		sshEnabled    bool
+		environmentSA string
+		orgAllowlist  []string
 	)
 	cmd := &cobra.Command{
-		Use:   "create [name] --hostname=[hostname] --cluster-address=[clusterAddress]",
+		Use:   "create [name]",
 		Args:  xcobra.ExactArgs(1),
 		Short: "create a new workspace provider.",
 		Long:  "Create a new Coder workspace provider.",
```
deansheather, line 54:
```suggestion
		Long:  `Create a new Coder workspace provider.

Creates a new Kubernetes workspace provider with the specified details. No resources are created or validated on the cluster.`,
```

deansheather (approved):

coder-sdk/workspace_providers.go
```diff
@@ -68,26 +68,55 @@ func (c *DefaultClient) WorkspaceProviders(ctx context.Context) (*WorkspaceProvi
...
+	EnvproxyAccessURL string                  `json:"envproxy_access_url"`
+	DevurlHost        string                  `json:"devurl_host"`
+	ErrorMessage      string                  `json:"error_message"`
+	CordonReason      string                  `json:"cordon_reason"`
+	EnableNetV2       bool                    `json:"enable_netv2"`
+}
+
+// WorkspaceProviderKubeConfig defines the kubernetes parameters of a workspace provider.
+type WorkspaceProviderKubeConfig struct {
+	ClusterCA           string `json:"cluster_ca" diff:"cluster_ca"`
```
jawnsy, line 106: what does the `diff` struct tag do?

internal/cmd/providers.go
jawnsy: Can we add a sentence about workspace providers here? just helps in case people don't already know what they are. Something like: "Create a workspace provider and connect it with your Coder installation. Workspace providers enable Coder to dynamically provision and decommission(?) developer workspaces in any Kubernetes cluster."

Not sure if "decommission" is the right word to use here...

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
-				return xerrors.Errorf("parse hostname: %w", err)
-			}
+			return nil
+		},
+	}
 
-			if cemanagerURL.Scheme != ingressHost.Scheme {
-				return xerrors.Errorf("Coder access url and hostname must have matching protocols: coder access url: %s, workspace provider hostname: %s", cemanagerURL.String(), ingressHost.String())
-			}
+	cmd.Flags().BoolVar(&fromContext, "from-context", false, "use current kube context to retrieve cluster-cert, cluster-address, sa-token, and namespace if not set explicitly")
```
jawnsy, line 113: can we match the config name used by Kubernetes (kubectl or helm) when possible

```
$ kubectl options
The following options can be passed to any command:

      --cluster='': The name of the kubeconfig cluster to use
      --context='': The name of the kubeconfig context to use
      --kubeconfig='': Path to the kubeconfig file to use for CLI requests.
  -n, --namespace='': If present, the namespace scope for this CLI request
used to contact the server is used
      --token='': Bearer token for authentication to the API server
```

or Helm, which uses `--kube-context`

```
$ helm install --help
      --kube-apiserver string       the address and the port for the Kubernetes API server
      --kube-as-group stringArray   group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string         username to impersonate for the operation
      --kube-ca-file string         the certificate authority file for the Kubernetes API server connection
      --kube-context string         name of the kubeconfig context to use
      --kube-token string           bearer token used for authentication
      --kubeconfig string           path to the kubeconfig file
  -n, --namespace string            namespace scope for this request
```

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
+  Name:  %s
+Subjects:
+  - Kind: ServiceAccount
+    Name: %s
+    Namespace: %s
+
+`, xkube.ResourceName, xkube.ResourceName, kctx.Namespace)
+
+			if !skipPrompt {
+				fmt.Println("Press the Enter Key to continue")
```
jawnsy, line 183: a y/n confirmation might be better, because it makes it easy to exit (though people can also do ^C)

internal/x/xkube/provider.go
jawnsy: Yeah, one of the distinctions between labels and annotations is that [annotations can't be searched](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/):

> You can use either labels or annotations to attach metadata to Kubernetes objects. Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects. The metadata in an annotation can be small or large, structured or unstructured, and can include characters not permitted by labels.

internal/cmd/providers.go
```diff
@@ -36,100 +37,166 @@ func providersCmd() *cobra.Command {
...
+		Use:   "create [name]",
 		Args:  xcobra.ExactArgs(1),
 		Short: "create a new workspace provider.",
 		Long:  "Create a new Coder workspace provider.",
-		Example: `# create a new workspace provider in a pending state
+		Example: `# create a new workspace provider
 
-coder providers create my-provider --hostname=https://provider.example.com --cluster-address=https://255.255.255.255`,
+coder providers create my-provider`,
 		RunE: func(cmd *cobra.Command, args []string) error {
```
jawnsy, line 58: Is it possible to add a "dry run" mode which can dump all the resulting YAML files? That way, someone can send the configs to their k8s admin to run, if they don't have the necessary cluster permissions. Kind of like `helm template` does

jawnsy (approved):

f0ssel:
Closing in favor of a dashboard only flow that turned out way nicer than expected

PR #361, opened 2021-05-26
Adjust HTTPError to read response body before it is closed

nathanpotter:
Resolves:
https://app.clubhouse.io/coder/story/12938/remove-http-response-from-httperror-in-coder-sdk

### DIFF
--- a/coder-sdk/error.go
+++ b/coder-sdk/error.go
@@ -1,8 +1,10 @@
 package coder
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 
 	"golang.org/x/xerrors"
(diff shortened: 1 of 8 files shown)


coder-sdk/error.go
```diff
@@ -36,7 +36,7 @@ type APIErrorMsg struct {
 // an *APIError.
 func NewHTTPError(resp *http.Response) *HTTPError {
 	var buf bytes.Buffer
-	_, err := io.CopyN(&buf, resp.Body, 1<<20)
+	_, err := io.Copy(&buf, resp.Body)
```
ammario, line 39: 👍 nice simplification. I can't think of a case where the error body would be excessively large.

ammario (approved):
Nice bug de-risk

PR #364, opened 2021-06-04
wsnet dial policy

deansheather:
Adds `Policies` field to `ProtoMessage` which denotes which network + host + port combinations a client connection may use. Makes `ProtoMessage` public.

### DIFF
--- a/wsnet/listen.go
+++ b/wsnet/listen.go
@@ -21,6 +21,24 @@ const OpNotPermittedByPolicy = "port_not_permitted"
 
 var connectionRetryInterval = time.Second
 
+// Codes for DialChannelResponse.
+const (
+	CodeDialErr       = "dial_error"
+	CodePermissionErr = "permission_error"
+	CodeBadAddressErr = "bad_address_error"
+)
+
+// DialChannelResponse is used to notify a dial channel of a
+// listening state. Modeled after net.OpError, and marshalled
+// to that if Net is not "".
+type DialChannelResponse struct {
+	Code string
+	Err  string
+	// Fields are set if the code is CodeDialErr.
+	Net string
+	Op  string
+}
+
 // Listen connects to the broker proxies connections to the local net.
 // Close will end all RTC connections.
 func Listen(ctx context.Context, broker string) (io.Closer, error) {
@@ -265,7 +283,7 @@ func (l *listener) handle(msg BrokerMessage) func(dc *webrtc.DataChannel) {
 				return
 			}
 
-			var init dialChannelMessage
+			var init DialChannelResponse
 			sendInitMessage := func() {
 				initData, err := json.Marshal(&init)
 				if err != nil {
@@ -285,10 +303,11 @@ func (l *listener) handle(msg BrokerMessage) func(dc *webrtc.DataChannel) {
 
 			network, addr, err := msg.getAddress(dc.Protocol())
 			if err != nil {
+				init.Code = CodeBadAddressErr
 				init.Err = err.Error()
 				var policyErr notPermittedByPolicyErr
 				if errors.As(err, &policyErr) {
-					init.Op = OpNotPermittedByPolicy
+					init.Code = CodePermissionErr
 				}
 				sendInitMessage()
 				return
@@ -296,6 +315,7 @@ func (l *listener) handle(msg BrokerMessage) func(dc *webrtc.DataChannel) {
 
 			conn, err := net.Dial(network, addr)
 			if err != nil {
+				init.Code = CodeDialErr
 				init.Err = err.Error()
 				if op, ok := err.(*net.OpError); ok {
 					init.Net = op.Net
--- a/wsnet/listen.go
+++ b/wsnet/listen.go
@@ -17,12 +17,9 @@ import (
 	"cdr.dev/coder-cli/coder-sdk"
 )
 
-var connectionRetryInterval = time.Second
+const OpNotPermittedByPolicy = "port_not_permitted"
 
-var localNet = &net.IPNet{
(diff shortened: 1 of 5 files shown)


wsnet/proto.go
```diff
@@ -4,18 +4,33 @@ import (
...
 
-// protoMessage is used for brokering a dialer and listener.
+// DialPolicy a single network + address + port combinations that a connection
+// is permitted to use.
+type DialPolicy struct {
+	// If network is empty, it applies to all networks.
+	Network string `json:"network"`
+	// If host is empty, it applies to all hosts. "localhost" and any IP address
+	// under "127.0.0.0/8" can be used interchangeably.
+	Host string `json:"address"`
```
kylecarbs, line 14: Would `Address` make more sense here? I'm concerned host is a bit ambiguous as to whether you should include the port here or not.

wsnet/proto.go
```diff
@@ -4,18 +4,33 @@ import (
...
+// ProtoMessage is used for brokering a dialer and listener.
 //
 // Dialers initiate an exchange by providing an Offer,
 // along with a list of ICE servers for the listener to
 // peer with.
 //
 // The listener should respond with an offer, then both
 // sides can begin exchanging candidates.
-type protoMessage struct {
+type ProtoMessage struct {
```
kylecarbs, line 27: Thoughts on calling this a `BrokerMessage` instead?

wsnet/listen.go
```diff
@@ -316,3 +336,58 @@ func (l *listener) Close() error {
...
+	if localNet.Contains(ip) {
+		return "localhost"
+	}
+	return addr
+}
+
+// getAddress parses the data channel's protocol into an address suitable for
+// net.Dial. It also verifies that the ProtoMessage permits connecting to said
+// address.
+func getAddress(msg ProtoMessage, protocol string) (netwk, addr string, err error) {
```
kylecarbs, line 356: Might make more sense to make this a function on `ProtoMessage`... something like:

```go
func (p ProtoMessage) Allowed(net, addr string) (bool, error) {

}
```

Then we can test it really well too... this is definitely a sensitive piece of code we wanna make sure is :bricks:

wsnet/listen.go
```diff
@@ -316,3 +336,58 @@ func (l *listener) Close() error {
 func (l *listener) Addr() net.Addr {
 	return nil
 }
+
+// normalizeHost converts all representations of "localhost" to "localhost".
+func normalizeHost(addr string) string {
```
kylecarbs, line 341: Calling `net.LookupIP` might be a cleaner way to standardize the URLs.

wsnet/listen.go
```diff
@@ -316,3 +336,58 @@ func (l *listener) Close() error {
...
+			continue
+		}
+		if p.Port != 0 && p.Port != portParsedU16 {
+			continue
+		}
+
+		return network, fullAddr, nil
+	}
+
+	return "", "", fmt.Errorf("connections are not permitted to %q", err)
```
kylecarbs, line 392: It'd be nice if we had this as an exposed Err... exactly like we do for the `Op` err type.

Something like `ErrNoPermission`

kylecarbs (changes requested):

wsnet/proto.go
deansheather: Host means IP or hostname only I'll make it clear in the comment

wsnet/listen.go
deansheather: Great idea!

wsnet/listen.go
deansheather: I don't want to make DNS requests from this. If people want to do `localhost.example.com` then their connections will just get aborted. They can just use `localhost` or `127.0.0.1` instead.

wsnet/listen.go
deansheather: did this as `func (msg BrokerMessage) getAddress(protocol string) (netwk, addr string, err error)` instead

wsnet/listen.go
deansheather: added a new type `notPermittedByPolicyErr` and a op `OpNotPermittedByPolicy = "port_not_permitted"`

wsnet/listen.go
```diff
@@ -18,6 +17,8 @@ import (
 	"cdr.dev/coder-cli/coder-sdk"
 )
 
+const OpNotPermittedByPolicy = "port_not_permitted"
```
kylecarbs, line 20: I did a bad job at explaining my thoughts for this... but it'd be nice if we added a new error type instead. Primarily so we could do `xerrors.Is` to check if perms are denied.

I'm hesitant to make this a new networking Op, since this isn't really a networking error... more of a protocol error.

kylecarbs (approved):
Minor-ish nit, but I'll approve for future ease!

wsnet/listen.go
kylecarbs: Fair enough fair enough.

wsnet/listen.go
kylecarbs: Can be removed

wsnet/listen.go
```diff
@@ -18,8 +17,28 @@ import (
...
+
 var connectionRetryInterval = time.Second
 
+// Codes for DialChannelResponse.
+const (
+	CodeDialErr       = "dial_error"
+	CodePermissionErr = "permission_error"
+	CodeBadAddressErr = "bad_address_error"
```
kylecarbs, line 28: Do we need to export these codes? Shouldn't we do `xerrors.Is` instead?

kylecarbs (approved):

PR #365, opened 2021-06-08
feat: add TLS commands

lilshoff:


internal/cmd/tls.go
```diff
@@ -0,0 +1,255 @@
...
+			if !resp.Success {
+				return xerrors.New("Failed to generate certificate via Let's Encrypt")
+			}
+
+			clog.LogSuccess("Generating certificate via Let's Encrypt. This may take a few minutes to complete.")
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVarP(&showProviderInfo, "info", "i", false, "Show supported DNS providers and required credentials for each")
```
jawnsy, line 207: I would suggest avoiding the short forms (`i`) as it makes things easier to mix up, and we can always add short forms later, but cannot remove them without breaking backward compatibility

internal/cmd/tls.go
```diff
@@ -0,0 +1,255 @@
...
+		domains     []string
+		dnsProvider string
+		credentials map[string]string
+	)
+	cmd := &cobra.Command{
+		Use:   "acme",
+		Short: "Generate certificate via Let's Encrypt",
+		Example: `
+tls acme --info
+tls acme --email me@example.com --domains a.example.com --domains b.example.com --provider route53 --credentials AWS_ACCESS_KEY_ID=your-key-id --credentials AWS_SECRET_ACCESS_KEY=your-secret-key --credentials AWS_REGION=your-region`,
```
jawnsy, line 148: it would be nice if we could pass credentials from environment variables if possible, for example: `tls acme --credentials AWS_ACCESS_KEY_ID` would read the `AWS_ACCESS_KEY_ID` environment variable and use that as a credential (or event do that automatically if those variables are set). This is because AWS users often have these environment variables set, and they are respected by tools they're using.

I think if you set the env vars, then the `aws cli` command will use them: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html - and it would be nice if we could do that with this tool too, like if someone specifies route53, then use their default env vars if set.

Feel free to do that in a separate PR (or not at all) though! Not a big deal, just a UX enhancement

jawnsy (approved):

internal/cmd/tls.go
```diff
@@ -0,0 +1,255 @@
...
+			if !resp.Success {
+				return xerrors.New("Failed to apply custom TLS certificate")
+			}
+
+			clog.LogSuccess("Uploaded custom TLS certificate")
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&certFilepath, "cert", "", "", "Full path to a PEM encoded certificate file")
```
jawnsy, line 128: The P-suffixed methods are for the short flags, and I think you can use StringVar() if there's no short form. From the [docs for pflag](https://pkg.go.dev/github.com/spf13/pflag):

>The pflag package also defines some new functions that are not in flag, that give one-letter shorthands for flags. You can use these by appending 'P' to the name of any function that defines a flag.

internal/cmd/tls.go
```diff
@@ -204,12 +204,12 @@ tls acme --email me@example.com --domains a.example.com --domains b.example.com
...
 
-	cmd.Flags().BoolVarP(&showProviderInfo, "info", "", false, "Show supported DNS providers and required credentials for each")
-	cmd.Flags().BoolVarP(&agreeTOS, "agree-tos", "", false, "Agree to ACME Terms of Service")
-	cmd.Flags().StringVarP(&email, "email", "e", "", "Email to use for ACME account")
-	cmd.Flags().StringArrayVarP(&domains, "domains", "", []string{}, "Domains to request certificates for")
-	cmd.Flags().StringVarP(&dnsProvider, "provider", "", "", "DNS provider hosting your domains")
-	cmd.Flags().StringToStringVarP(&credentials, "credentials", "c", map[string]string{}, "DNS provider credentials")
+	cmd.Flags().BoolVar(&showProviderInfo, "info", false, "Show supported DNS providers and required credentials for each")
+	cmd.Flags().BoolVar(&agreeTOS, "agree-tos", false, "Agree to ACME Terms of Service")
+	cmd.Flags().StringVar(&email, "email", "e", "Email to use for ACME account")
```
jawnsy, line 209: probably removed the wrong parameter here? This sets the default to "e"
```suggestion
	cmd.Flags().StringVar(&email, "email", "", "Email to use for ACME account")
```

docs/coder_tls_acme.md
```diff
@@ -18,9 +18,9 @@ tls acme --email me@example.com --domains a.example.com --domains b.example.com
 
 ```
       --agree-tos                    Agree to ACME Terms of Service
-  -c, --credentials stringToString   DNS provider credentials (default [])
+      --credentials stringToString   DNS provider credentials (default [])
       --domains stringArray          Domains to request certificates for
```
jawnsy, line 22: How does pflag accept arrays? Are they specified like "[coder.com,abc.com]" or some other way?

jawnsy (approved):

docs/coder_tls_acme.md
lilshoff: You have to specify them one by one (`--domains a --domains b`)
It took me a while to figure that out, so I provide an example for the `tls acme` command


`coder tls acme -h` shows

```tls acme --email me@example.com --domains a.example.com --domains b.example.com --provider route53 --credentials AWS_ACCESS_KEY_ID=your-key-id --credentials AWS_SECRET_ACCESS_KEY=your-secret-key --credentials AWS_REGION=your-region```

jawnsy: Ahh I see. Do you mind adding a similar example for domains?

jawnsy: Ah nevermind, your example is for the full subcommand, alright then, carry on!

PR #366, opened 2021-06-09
feat: support setting policy template via local file or restoring to default

lilshoff:


internal/cmd/workspaces.go
```diff
@@ -752,3 +753,76 @@ func buildUpdateReq(ctx context.Context, client coder.Client, conf updateConf) (
...
+				} else {
+					fmt.Printf("Workspace %q in organization %q:\n", workspace.Name, workspace.OrganizationID)
+				}
+
+				fmt.Println(mc.String())
+			}
+			return nil
+		},
+	}
+	cmd.Flags().BoolVarP(&dryRun, "dry-run", "", false, "skip setting policy template, but view errors/warnings about how this policy template would impact existing workspaces")
```
lilshoff, line 822: Considering we only support "site" template now, I didn't make a flag for "scope" or anything. 

We'll need it if/when we support "org" and "site" templates... but I'm wondering if there's any benefit to adding a flag for it now?

sreya:
Similar to my comments in the other PR I would avoid adding this until we have proper Github App support.

lilshoff:
OK so do we want to (1) remove support for remote tracked policy templates or (2) not expose setting policy templates in the CLI at all?

coder-sdk/workspace.go
```diff
@@ -364,3 +366,99 @@ func (c *DefaultClient) WorkspacesByWorkspaceProvider(ctx context.Context, wpID
...
+		sb.WriteString("If workspace is updated to the latest template:\n")
+		if len(mc.LatestTemplateWarnings) != 0 {
+			sb.WriteString(fmt.Sprintf("Warnings: \n%s\n", strings.Join(mc.LatestTemplateWarnings, "\n")))
+		}
+		if mc.LatestTemplateError != nil {
+			sb.WriteString(fmt.Sprintf("Errors: \n%s\n", strings.Join(mc.LatestTemplateError.Msgs, "\n")))
+		}
+	}
+
+	return sb.String()
```
Emyrk, line 435: Something that might be helpful is some sort of summary at the end. I have 200 workspaces in my dogfood, and it would be nice to see something like:

```
102 Workspaces will not be able to be rebuilt.
152 Workspaces have warnings but are still functional.
```

In the future, it would be great to merge-common errors and report which users are affected maybe? But for now I think a simple summary like that would be great.

internal/cmd/workspaces.go
```diff
@@ -752,3 +753,76 @@ func buildUpdateReq(ctx context.Context, client coder.Client, conf updateConf) (
...
+		ref      string
+		repo     string
+		filepath string
+		dryRun   bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "policy-template",
+		Short: "Set workspace policy template",
+		Long:  "Set workspace policy template",
```
Emyrk, line 768: Can we put a note only site admins can do this?

internal/cmd/workspaces.go
Emyrk: You can add the scope, make `site` the default? I'm good with either decision.

internal/cmd/workspaces.go
```diff
@@ -752,3 +753,76 @@ func buildUpdateReq(ctx context.Context, client coder.Client, conf updateConf) (
...
+				return handleAPIError(err)
+			}
+
+			resp, err := client.SetPolicyTemplate(ctx, version.TemplateID, coder.TemplateScopeSite, dryRun)
+			if err != nil {
+				return handleAPIError(err)
+			}
+
+			for _, mc := range resp.MergeConflicts {
+				workspace, err := client.WorkspaceByID(ctx, mc.WorkspaceID)
```
Emyrk, line 806: I like the extra details. In my dogfood of 200 workspaces this was still pretty quick :+1:

Emyrk (approved):

internal/cmd/workspaces.go
lilshoff: ✅

Emyrk:
> OK so do we want to (1) remove support for remote tracked policy templates or (2) not expose setting policy templates in the CLI at all?

Ah saw John's comment yea. Drop setting git tracked tracked templates.
I'm indifferent on if we should have a cli for it. You can always hide the command since it's already done 🤷

coder-sdk/workspace.go
lilshoff: I added a summary at the end.

Mine isn't too exciting bc I regularly delete my workspaces in my dev deployment, so I only have like 2.
I'm curious to see what you think of the summary when there are 200 🤔

lilshoff:
Yeah I initially didn't think this belonged in the CLI, but Ketan made a comment in [the tls frontend CH ticket](https://app.clubhouse.io/coder/story/13521/frontend-support-for-coderd-tls-management) "_I think having the loose goal of CLI feature parity with the graphical web UI is a good default position._"

Then there's also [this ticket](https://app.clubhouse.io/coder/story/11610/site-template-dryrun-to-see-affected-workspaces) which seems to imply we want this in the CLI

coder-sdk/workspace.go
```diff
@@ -364,3 +366,146 @@ func (c *DefaultClient) WorkspacesByWorkspaceProvider(ctx context.Context, wpID
...
+	updateConflicts := len(mc.LatestTemplateWarnings) != 0 || mc.LatestTemplateError != nil
+
+	if !currentConflicts && !updateConflicts {
+		sb.WriteString("No workspace conflicts\n")
+		return sb.String()
+	}
+
+	if currentConflicts {
+		if len(mc.CurrentTemplateWarnings) != 0 {
+			sb.WriteString(fmt.Sprintf("Warnings: \n%s\n", strings.Join(mc.CurrentTemplateWarnings, "\n")))
```
jawnsy, line 418: suggest using `fmt.Fprintf(&sb` here https://play.golang.org/p/21-gij80d3c

coder-sdk/workspace.go
jawnsy: I'm very curious about why @Emyrk 200 workspaces :joy:

Emyrk: Ha, my deployment has 200 workspaces because I was testing the response times of our apis at those numbers. I made a little deployment branch to make them fast, but I never made anything to delete them in bulk haha.

I figure I'll just keep em as a bit of a "stress test" for the UI

Emyrk: The summary at 200 is not the most helpful thing, but I'm not worried about optimizing for 200 workspaces right now. Just thought a summary would be nice even for like ~30 workspaces which is reasonable at a company

PR #369, opened 2021-06-18
feat: open coder workspace IDE

lilshoff:


internal/cmd/open.go
```diff
@@ -0,0 +1,70 @@
...
+				if w.Name == workspaceName {
+					workspaceID = w.ID
+				}
+			}
+			if workspaceID == "" {
+				return xerrors.Errorf("Workspace %q not found", workspaceName)
+			}
+
+			u := client.BaseURL()
+			u.Path += "/app/code"
```
jawnsy, line 52: Ah, I guess this only works with VSCode and not JetBrains etc? I guess if we maintain this pattern in the future, we could add an --app flag to control the exact URL we use?

internal/cmd/open.go
```diff
@@ -0,0 +1,70 @@
...
+			client, err := newClient(ctx, true)
+			if err != nil {
+				return err
+			}
+
+			if workspaceName == "" {
+				return xerrors.Errorf("Missing required parameter: --workspace")
+			}
+
+			workspaces, err := getWorkspaces(ctx, client, user)
```
jawnsy, line 36: If the workspace isn't running, should we start it? Or should we check the status and emit an error message if it's off?

lilshoff: I like the idea of just starting it for them. 
I assume if they're running "coder workspace open" they either assume its already built or ultimately _do_ want it built, so we might as well just automate that step

internal/cmd/workspaces.go
```diff
@@ -752,3 +754,71 @@ func buildUpdateReq(ctx context.Context, client coder.Client, conf updateConf) (
...
+	var (
+		workspaceName string
+	)
+
+	cmd := &cobra.Command{
+		Use:     "open",
+		Short:   "Open Coder workspace IDE",
+		Long:    "Open Coder workspace IDE in your default browser",
+		Example: "coder workspace open --name dev",
+		RunE: func(cmd *cobra.Command, args []string) error {
```
kylecarbs, line 768: We can add the ability to open the workspace with VS Code Remote locally.

1. Run `config-ssh` for them.
2. Ensure VS Code is installed.
3. Launch `vscode://vscode-remote/ssh-remote+coder.rem` scheme. (replace coder.rem with the target instance).

internal/cmd/workspaces.go
```diff
@@ -752,3 +754,71 @@ func buildUpdateReq(ctx context.Context, client coder.Client, conf updateConf) (
...
+func openWorkspaceCmd() *cobra.Command {
+	var (
+		workspaceName string
+	)
+
+	cmd := &cobra.Command{
+		Use:     "open",
+		Short:   "Open Coder workspace IDE",
+		Long:    "Open Coder workspace IDE in your default browser",
+		Example: "coder workspace open --name dev",
```
kylecarbs, line 767: Should name be an argument instead of a flag? Maybe both?

Sounds great to do `coder workspace open dev`.

kylecarbs:
It'd be more powerful if this was focused on the local experience eg. VS Code Remote SSH instead of browser. If a user can obtain the CLI locally, most likely they can run VS Code locally.

PR #378, opened 2021-07-07
chore: Cache SSH connections

kylecarbs:
Speedy speedy speed...

Works for Linux and Darwin systems, [but not Windows](https://github.com/microsoft/vscode-remote-release/issues/629#issuecomment-503650294).

https://user-images.githubusercontent.com/7122116/124811861-8e0f8600-df28-11eb-8925-51a301e5ff54.mp4

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/cmd/configssh.go
```diff
@@ -208,7 +209,7 @@ func makeNewConfigs(binPath, userName string, workspaces []coderutil.WorkspaceWi
 
 func makeSSHConfig(binPath, host, userName, workspaceName, privateKeyFilepath string, tunnel bool) string {
 	if tunnel {
-		return fmt.Sprintf(
+		host := fmt.Sprintf(
```
jawnsy, line 212: Is it possible/straightforward to add an integration test for this?

jawnsy (approved):
Nice work! Does this also multiplex connections over a single TCP stream, or was that a separate feature?

internal/cmd/configssh.go
kylecarbs: Good idea.... will do!

kylecarbs:
It multiplexes over a single TCP stream!

f0ssel (approved):

PR #381, opened 2021-07-09
feat: Add Closed func to Dialer

kylecarbs:


wsnet/dial_test.go
```diff
@@ -229,6 +229,24 @@ func TestDial(t *testing.T) {
...
+		}
+		dialer, err := DialWebsocket(context.Background(), connectAddr, nil)
+		if err != nil {
+			t.Error(err)
+			return
+		}
+		go func() {
+			_ = dialer.Close()
+		}()
+		<-dialer.Closed()
```
deansheather, line 248: do a select against a timer here so we can ensure it closes quickly

wsnet/dial.go
```diff
@@ -136,6 +140,10 @@ func (d *Dialer) negotiate() (err error) {
 				_ = connCloser.Close()
 			}
 			d.connClosers = make([]io.Closer, 0)
+			if !d.closed {
+				d.closed = true
+				close(d.closedChan)
+			}
```
deansheather, line 146: Instead of having a bool have a select with a default:
```go
select {
  case <-d.closedChan:
    return
  default:
}

close(d.closedChan)
```

deansheather (approved):

PR #382, opened 2021-07-09
fix: Close Ping DataChannel when connection ends

kylecarbs:
Previously, Ping() would hang forever due to the DataChannel
never closing when the RTC connection ended.


wsnet/dial_test.go
```diff
@@ -67,7 +69,38 @@ func TestDial(t *testing.T) {
 		}
 	})
 
+	t.Run("Ping Close", func(t *testing.T) {
```
deansheather, line 72:
```suggestion
	t.Run("PingClose", func(t *testing.T) {
```

deansheather (approved):

PR #384, opened 2021-07-13
feat: Enable TURN proxying over WebSocket

kylecarbs:
Exposes a `proxy.Dialer` to `wsnet.Dial`. Enables tunneling of the TURN server through an arbitrary protocol.

`TURNWebSocketDialer` exists to help convert a WebSocket connection.


wsnet/conn.go
```diff
@@ -50,7 +46,108 @@ func ConnectEndpoint(baseURL *url.URL, workspace, token string) string {
...
+		conn: websocket.NetConn(context.Background(), conn, websocket.MessageBinary),
+	}, nil
+}
+
+// turnProxyConn is a net.Conn wrapper that returns a TCPAddr for the
+// LocalAddr and RemoteAddr functions. pion/ice unsafely checks the
+// types. See: https://github.com/pion/ice/blob/e78f26fb435987420546c70369ade5d713beca39/gather.go#L448
+type turnProxyConn struct {
+	conn net.Conn
+}
```
deansheather, line 110: This could just be reimplemented as
```go
type turnProxyConn struct {
    net.Conn
}
```

or even just `type turnProxyConn net.Conn` and override the LocalAddr method

deansheather (approved):

wsnet/conn.go
kylecarbs: Ahh :brain:

PR #387, opened 2021-07-16
Add satellite commands

f0ssel:


internal/cmd/satellites.go
```diff
@@ -0,0 +1,217 @@
...
+
+func createSatelliteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [name] [satellite_access_url]",
+		Args:  xcobra.ExactArgs(2),
+		Short: "create a new satellite.",
+		Long:  "Create a new Coder satellite.",
+		Example: `# create a new satellite
+
+coder satellites create eu-west https://eu-west-coder.com`,
```
deansheather, line 52:
```suggestion
coder satellites create eu-west https://eu-west.coder.com`,
```

internal/cmd/satellites.go
```diff
@@ -0,0 +1,217 @@
...
+			fmt.Printf(`The following satellite will be created:
+Name: %s
+
+Public Key: 
+%s
+
+Fingerprint:
+%s
+
+Do you wish to continue? (y)
```
deansheather, line 107: Typically having an option in parenthesis means that it's the default if you hit enter and don't type anything, I don't think your code does that.

deansheather (approved):

kylecarbs (approved):
Awesome.

internal/cmd/satellites.go
f0ssel: Why? Part of the reason I picked that was to illustrate it could be a completely separate domain if desired.

internal/cmd/satellites.go
f0ssel: Should I do `(y/n)`? I still wanna prompt the user to hit `y`

internal/cmd/satellites.go
deansheather: Well we don't own `eu-west.coder.com` so we'd be displaying a domain name that anyone could buy.

internal/cmd/satellites.go
deansheather: Yeah that's good. In that case you should also `response = strings.ToLower(strings.TrimSpace(response))` the response before checking

internal/cmd/satellites.go
f0ssel: Ah fair

PR #388, opened 2021-07-16
Add logs to wsnet listener

deansheather:
TODO:
- [x] Add logger
- Log:
    - [x] Startup
    - [x] Handshake WS connect
    - [x] Handshake WS stream opened in yamux
    - [x] ICE candidates received (and contents)
    - [x] Sending ICE candidates (and contents)
    - [x] Connection state changes
    - [x] Timeouts/ping failures


f0ssel (approved):
Do we current run the agent with debug enabled? We should if not at least for initial release

PR #389, opened 2021-07-17
fix: Add TURN URL to handshake message for custom proxying URLs

kylecarbs:


wsnet/listen.go
```diff
@@ -200,7 +201,19 @@ func (l *listener) negotiate(conn net.Conn) {
...
+			var turnProxy proxy.Dialer
+			if msg.TURNProxyURL != "" {
+				u, err := url.Parse(msg.TURNProxyURL)
+				if err != nil {
+					closeError(fmt.Errorf("parse turn proxy url: %w", err))
+					return
+				}
+				turnProxy = &turnProxyDialer{
+					baseURL: u,
+					token:   l.turnProxyAuthToken,
```
deansheather, line 213: There's no validation here. Any client could send any URL (like `wss://malicious.example.com`) and the listener would send it's token to that URL. We can solve this by:
1. MITMing the handshake on coderd and validating that the URL is trusted, which would require storing satellite URLs in the DB which we don't do right now
2. Having the client send the token as part of the URL in the handshake payload
    - If it's your own workspace that's fine
    - For satellite/coderd proxied requests it can be done by sending the envagent token which the listener already has
    - For connecting directly to someone else's workspace (which we don't support right now but may in the future), it's not fine

We should just do option 2 for now since it covers our current usecase

deansheather (changes requested):

wsnet/listen.go
kylecarbs: The client would already have the token, right? If they leaked to a malicious URL I believe it'd have to be intentional.

f0ssel: Since we already manage the setup of a satellite through the cli and that's pretty simple, adding a url field to the db table isn't the worst imo. I can't comment on if this PR is secure or not really, but from the proposed solutions I think 1 seems fine for now and is simplest. If they change it down the road and don't update it with the cli, we'll spot 403s back at them and it'll be pretty obvious to debug.

f0ssel (approved):

wsnet/listen.go
deansheather: Closing since this isn't a security problem until we let other people connect to other workspaces

deansheather (approved):

PR #391, opened 2021-07-20
feat: Add DialCache for key-based connection caching

kylecarbs:


wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+
+type DialerCache struct {
+	dialerFunc  dialerFunc
+	ttl         time.Duration
+	flightGroup *singleflight.Group
+
+	closed  chan struct{}
+	mut     sync.RWMutex
+	dialers map[string]*Dialer
+	atime   map[string]time.Time
```
deansheather, line 38: You should use a struct instead of having two maps for the same key.

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+		dialer := dialer
+		go func() {
+			defer wg.Done()
+
+			evict := false
+			select {
+			case <-dialer.Closed():
+				evict = true
+			default:
+			}
```
deansheather, line 71:
```suggestion
```

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Close closes all cached dialers.
+func (d *DialerCache) Close() error {
+	d.mut.Lock()
+	defer d.mut.Unlock()
+
+	for key, dialer := range d.dialers {
+		d.flightGroup.Forget(key)
+
```
deansheather, line 144:
```suggestion
```

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+		err := dialer.Close()
+		if err != nil {
+			return err
+		}
+	}
+	close(d.closed)
```
deansheather, line 150: You should check if it's closed at the top of Dial() and after the flightGroup.Do

wsnet/cache_test.go
```diff
@@ -0,0 +1,67 @@
...
+		require.NoError(t, err)
+		defer l.Close()
+
+		cache := DialCache(time.Hour, func(ctx context.Context, key string) (*Dialer, error) {
+			return DialWebsocket(ctx, connectAddr, nil)
+		})
+		_, cached, err := cache.Dial(context.Background(), "example")
+		require.NoError(t, err)
+		require.Equal(t, cached, false)
+		_, cached, err = cache.Dial(context.Background(), "example")
```
deansheather: Check that the returned dialers are `assert.Same()`

wsnet/cache_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+		cache := DialCache(time.Hour, func(ctx context.Context, key string) (*Dialer, error) {
+			return DialWebsocket(ctx, connectAddr, nil)
+		})
+
+		conn, cached, err := cache.Dial(context.Background(), "example")
+		require.NoError(t, err)
+		require.Equal(t, cached, false)
+		require.NoError(t, conn.Close())
+		_, cached, err = cache.Dial(context.Background(), "example")
```
deansheather: Check that they aren't the `assert.Same()`

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+			if !evict {
+				ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+				defer cancel()
+				err := dialer.Ping(ctx)
+				if err != nil {
+					evict = true
+				}
+			}
+
+			if evict {
```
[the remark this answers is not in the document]
kylecarbs: Huh?

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+				_ = dialer.Close()
+				d.mut.Lock()
+				delete(d.atime, key)
+				delete(d.dialers, key)
+				d.mut.Unlock()
```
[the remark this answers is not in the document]
kylecarbs: I don't _think_ so. This is inside a WaitGroup goroutine.

wsnet/cache.go
```diff
@@ -0,0 +1,154 @@
...
+				if err != nil {
+					evict = true
+				}
+			}
+
+			if evict {
+				_ = dialer.Close()
+				d.mut.Lock()
+				delete(d.atime, key)
+				delete(d.dialers, key)
```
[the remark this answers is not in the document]
kylecarbs: Ahh interesting. How should I do that?

wsnet/cache.go
kylecarbs: Why remove this? We want to evict closed connections.

wsnet/cache.go
kylecarbs: I felt like it was a bit overkill here... I'll add a comment denoting they're the same.

wsnet/cache.go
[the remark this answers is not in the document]
deansheather: Use `else` instead of two opposite `if`s

wsnet/cache.go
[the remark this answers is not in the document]
deansheather: Check if `d.dialers[key]` exists, and is equal to the current dialer pointer you have, if it isn't then don't delete anything

wsnet/cache.go
deansheather: oops I thought this was d.Closed my bad

wsnet/cache.go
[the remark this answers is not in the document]
deansheather: This is the last statement in the goroutine so it's effectively the same. You get protection from mut getting left locked from panics if you use defer, so it's better to use defer wherever possible

wsnet/cache.go
```diff
@@ -0,0 +1,163 @@
...
+	return dc
+}
+
+type DialerCache struct {
+	ttl         time.Duration
+	flightGroup *singleflight.Group
+	closed      chan struct{}
+	mut         sync.RWMutex
+
+	// Key is the "key" of a dialer.
```
deansheather, line 35:
```suggestion
	// Key is the "key" of a dialer, which is usually the workspace ID.
```

wsnet/cache_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		cache := DialCache(0)
+
+		_, cached, err := cache.Dial(context.Background(), "example", dialFunc(connectAddr))
+		require.NoError(t, err)
+		require.Equal(t, cached, false)
+		cache.evict()
+		_, cached, err = cache.Dial(context.Background(), "example", dialFunc(connectAddr))
+		require.NoError(t, err)
+		require.Equal(t, cached, false)
```
deansheather, line 68: Check NotSame here too

deansheather (approved):

wsnet/cache.go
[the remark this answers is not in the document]
kylecarbs: Ahh the other one is to ping only if can't evict.

PR #394, opened 2021-07-21
fix: Add context to negotiate RTC func

kylecarbs:


wsnet/dial.go
```diff
@@ -173,6 +173,19 @@ func (d *Dialer) negotiate() (err error) {
...
+	go func() {
+		// If a connection is opened but the other end may not be, negotiation
+		// can get stuck forever. We don't want this, so we must listen to the
+		// context as well.
+		<-ctx.Done()
+		select {
+		case <-errCh:
+		default:
+			errCh <- ctx.Err()
+			close(errCh)
```
deansheather, line 185: You're closing this channel in multiple goroutines, it's going to panic on the second close attempt. You should consider not closing it, or making some `sendErr` function that wraps these two lines in a sync.Once

deansheather (approved):

wsnet/dial.go
kylecarbs: This switch checks if the channel is open... so it shouldn't close twice.

kylecarbs: Err I guess it would in the other one. I'll just check if it's open there too.

wsnet/rtc.go
```diff
@@ -242,8 +242,12 @@ func waitForConnectionOpen(ctx context.Context, conn *webrtc.PeerConnection) err
 	if conn.ConnectionState() == webrtc.PeerConnectionStateConnected {
 		return nil
 	}
-	ctx, cancelFunc := context.WithTimeout(ctx, time.Second*15)
+	ctx, cancelFunc := context.WithCancel(ctx)
 	defer cancelFunc()
+	if _, deadlineSet := ctx.Deadline(); !deadlineSet {
+		ctx, cancelFunc = context.WithTimeout(ctx, time.Second*15)
+		defer cancelFunc()
+	}
```
deansheather, line 250: nit
```suggestion
	var cancel context.CancelFunc
	if _, deadlineSet := ctx.Deadline(); deadlineSet {
		ctx, cancel = context.WithCancel(ctx)
	} else {
		ctx, cancel = context.WithTimeout(ctx, time.Second*15)
	}
	defer cancel()
```

deansheather (approved):

PR #395, opened 2021-07-22
fix: Close conn if context deadline is exceeded

kylecarbs:
Previously the error occurred but never propogated due to blocking fromm
the read loop in negotiate. Now, connection timeouts can properly occur.

### DIFF
--- a/wsnet/dial_test.go
+++ b/wsnet/dial_test.go
@@ -59,8 +59,9 @@ func TestDial(t *testing.T) {
 
 		ctx, cancelFunc := context.WithTimeout(context.Background(), time.Millisecond*50)
 		defer cancelFunc()
-		_, err := DialWebsocket(ctx, connectAddr, nil, nil)
+		dialer, err := DialWebsocket(ctx, connectAddr, nil, nil)
 		require.True(t, errors.Is(err, context.DeadlineExceeded))
+		require.Error(t, dialer.conn.Close(), "already wrote close")
 	})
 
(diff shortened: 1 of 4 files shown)


wsnet/dial_test.go
```diff
@@ -51,6 +52,17 @@ func ExampleDial_basic() {
...
 
 func TestDial(t *testing.T) {
+	t.Run("Timeout", func(t *testing.T) {
+		t.Parallel()
+
+		connectAddr, _ := createDumbBroker(t)
+
+		ctx, cancelFunc := context.WithTimeout(context.Background(), time.Millisecond*50)
+		defer cancelFunc()
+		_, err := DialWebsocket(ctx, connectAddr, nil, nil)
```
deansheather, line 62: Check underlying conn closed

wsnet/wsnet_test.go
```diff
@@ -68,7 +68,7 @@ func createDumbBroker(t testing.TB) (connectAddr string, listenAddr string) {
 		mut.Lock()
 		defer mut.Unlock()
 		if sess == nil {
-			t.Error("listen not called")
+			_, _ = io.Copy(io.Discard, nc)
```
deansheather, line 71: ??

deansheather (approved):

wsnet/wsnet_test.go
kylecarbs: I'll add a comment. This is to emulate a PubSub style scenario where we don't know if anyone's listening.

PR #401, opened 2021-07-26
Update last connection time on tunnel connections

f0ssel:


deansheather (approved):

coder-sdk/agent.go
```diff
@@ -0,0 +1,22 @@
+package coder
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// UpdateLastConnectionAt updates the last connection at attribute of a workspace.
+func (c *DefaultClient) UpdateLastConnectionAt(ctx context.Context, workspaceID string) error {
```
kylecarbs, line 10: We _technically_ already do this... right?

https://github.com/cdr/m/commit/39ee8c732aebcfaef1cc43c8f6392c5849a2ade9#diff-658b44e208cfee6825cc7eead08c6383fded8ecf6a6708091665c29663732a93R122

f0ssel: That only does so on connection, this will do so continuously for SSH connections and such

PR #408, opened 2021-08-03
Use webrtc for coder sh

Emyrk:
# What this does

Uses webrtc for `coder sh`

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -251,12 +251,12 @@ func makeSSHConfig(binPath, workspaceName, privateKeyFilepath string) string {
 	entry := fmt.Sprintf(
 		`Host coder.%s
    HostName coder.%s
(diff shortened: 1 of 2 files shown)


f0ssel (approved):

Emyrk:
@deansheather @f0ssel This would likely break any v1 providers. I'm guessing we already released the CLI for the last release so I can merge this yes?

f0ssel:
Yup this will be save for 1.22

kylecarbs (changes requested):
It might be cleaner to move our options into a map. That way, we could use them for `config-ssh` along with `coder sh`.

Emyrk:
@kylecarbs That sort of solves things, but the quoting was also different for the two cases. 

A more general solution would nice, I didn't try and tackle that here.

kylecarbs (approved):
Try to merge master in first... I changed some stuff up with the options that will conflict here!

PR #410, opened 2021-08-09
feat: Enable arbitrary SSH options passed via config-ssh

kylecarbs:
[ch15898]

### DIFF
--- a/internal/cmd/configssh.go
+++ b/internal/cmd/configssh.go
@@ -34,6 +34,7 @@ const sshEndToken = "# ------------END-CODER-ENTERPRISE------------"
 func configSSHCmd() *cobra.Command {
 	var (
 		configpath string
+		noCache    = false
 		remove     = false
 	)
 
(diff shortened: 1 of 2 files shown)


jawnsy (approved):
LGTM!

kylecarbs:
@jawnsy I tagged you again since the scope changed up quite a bit!

jawnsy (approved):
I guess this means that someone might have:

ControlPath none
ControlPath auto

... or similar, and SSH takes the option that comes first? It seems a little confusing to me but if it functionally works then it seems good

kylecarbs:
Definitely a bit confusing... I'll add comment that gets appended after custom options explaining the functionality!

kylecarbs:
Check it out whenever you get a shot... I think it's good now!

internal/cmd/configssh.go
```diff
@@ -240,32 +242,41 @@ func makeNewConfigs(binPath string, workspaces []coderutil.WorkspaceWithWorkspac
...
-   ConnectTimeout=0
-   IdentitiesOnly yes
-   IdentityFile="%s"
-`, workspaceName, workspaceName, binPath, workspaceName, privateKeyFilepath)
+func makeSSHConfig(binPath, workspaceName, privateKeyFilepath string, additionalOptions []string) string {
+	// Custom user options come first to maximizessh customization.
+	options := []string{}
+	if len(additionalOptions) > 0 {
+		options = []string{
+			"# Custom options. Duplicated values will always prefer the first!",
```
jawnsy, line 257: I think it would be better if we kept a `map[string]string` of the name/value pairs, and then override them based on key... but ssh might have some peculiarities with the config format, like being case-insensitive

I think this is an improvement over the status quo so definitely think this is good now, we can always revisit later if we get customer feedback that this is confusing :)

jawnsy (approved):

internal/cmd/configssh.go
kylecarbs: I did this in an iteration, but didn't want the jank of splitting on " " or "="... I wasn't sure if there were additional edge-cases, and didn't wanna risk it!

That's definitely a cleaner approach though, so someday :pray:

PR #411, opened 2021-08-09
feat: Display error to user when SSHing into offline workspace

kylecarbs:
Very fun lil hack to display user output on an SSH connection to offline workspaces.

image

Here's what it looks like from VS Code:

image

### DIFF
--- a/internal/cmd/tunnel.go
+++ b/internal/cmd/tunnel.go
@@ -12,12 +12,15 @@ import (
 
 	"cdr.dev/slog"
 	"cdr.dev/slog/sloggers/sloghuman"
+	"github.com/fatih/color"
 	"github.com/pion/webrtc/v3"
 	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh"
 	"golang.org/x/xerrors"
 
 	"cdr.dev/coder-cli/coder-sdk"
 	"cdr.dev/coder-cli/internal/x/xcobra"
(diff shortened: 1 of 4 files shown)


kylecarbs:
[ch15894]

internal/cmd/tunnel.go
```diff
@@ -59,20 +62,34 @@ coder tunnel my-dev 3000 3000
...
 
-			var workspaceID string
-			for _, workspace := range workspaces {
-				if workspace.Name == args[0] {
-					workspaceID = workspace.ID
-					break
+			if workspace.LatestStat.ContainerStatus != coder.WorkspaceOn {
+				color.NoColor = false
+				notAvailableError := clog.Error("workspace not available",
+					fmt.Sprintf("current status: \"%s\"", workspace.LatestStat.ContainerStatus),
```
jawnsy, line 73:
```suggestion
					fmt.Sprintf("current status: %q", workspace.LatestStat.ContainerStatus),
```

:stuck_out_tongue:

jawnsy (approved):
nice! what was the error message before?

internal/cmd/tunnel.go
kylecarbs: hahah ty

kylecarbs:
A context deadline exceeded :face_with_head_bandage: It'd attempt to connect and exchange candidates but since a workspace wasn't listening nothing would happen!

PR #412, opened 2021-08-11
fix: Continue attempting connections no matter what

kylecarbs:
Found this edge-case while restarting my SaaS instance ~20 times.

### DIFF
--- a/wsnet/listen.go
+++ b/wsnet/listen.go
@@ -71,7 +71,7 @@ func Listen(ctx context.Context, log slog.Logger, broker string, turnProxyAuthTo
 			default:
 			}
 
-			if errors.Is(err, io.EOF) || errors.Is(err, yamux.ErrKeepAliveTimeout) {
+			if err != nil {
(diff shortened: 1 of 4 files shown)


kylecarbs:
Shoutout Dean for the wonderful logging that made finding this ezpz :rainbow:

wsnet/listen.go
```diff
@@ -143,7 +143,11 @@ func (l *listener) dial(ctx context.Context) (<-chan error, error) {
 		defer close(errCh)
 		for {
 			conn, err := session.Accept()
+			if errors.Is(err, io.EOF) {
```
deansheather, line 146: Does this not mean the yamux session is permanently borked?

deansheather (approved):

wsnet/listen.go
kylecarbs: Yea you're right... debugging the error we're not checking now.

deansheather (approved):

PR #414, opened 2021-08-11
chore: Add line returns to clog output

kylecarbs:
Occasionally SSH connections wouldn't return on newlines... this fixes it!

Before:
image

After:
image

### DIFF
--- a/pkg/clog/clog.go
+++ b/pkg/clog/clog.go
@@ -35,12 +35,11 @@ type CLIError struct {
 // String formats the CLI message for consumption by a human.
 func (m CLIMessage) String() string {
 	var str strings.Builder
-	str.WriteString(fmt.Sprintf("%s: %s\r\n",
+	fmt.Fprintf(&str, "%s: %s\r\n",
 		color.New(m.Color).Sprint(m.Level),
-		color.New(color.Bold).Sprint(m.Header)),
-	)
+		color.New(color.Bold).Sprint(m.Header))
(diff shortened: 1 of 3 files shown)


pkg/clog/clog.go
```diff
@@ -35,12 +35,12 @@ type CLIError struct {
 // String formats the CLI message for consumption by a human.
 func (m CLIMessage) String() string {
 	var str strings.Builder
-	str.WriteString(fmt.Sprintf("%s: %s\n",
+	str.WriteString(fmt.Sprintf("%s: %s\r\n",
```
jawnsy, line 38: Instead of doing WriteString with Sprintf, we can instead use Fprintf since strings.Builder implements io.Writer:

```go
package main

import (
	"fmt"
	"strings"
)

func main() {
	var str strings.Builder
	str.WriteString(fmt.Sprintf("%s: %s\r\n", "prefix", "test"))
	fmt.Printf("WriteString/Sprintf: %s", str.String())
	
	str.Reset()
	fmt.Fprintf(&str, "%s: %s\r\n", "prefix", "test message")
	fmt.Printf("Fprintf: %s", str.String())
}
```

This reduces memory allocations and also reduces the amount of code

Try it on the Playground: https://play.golang.org/p/Bw98WJLVF81

kylecarbs: Oh wonderful! Will change!

deansheather (approved):

PR #416, opened 2021-08-12
feat: Option to list all workspaces

goodspark:
From an admin POV, this will be helpful for various reasons.

### DIFF
--- a/internal/cmd/workspaces.go
+++ b/internal/cmd/workspaces.go
@@ -70,6 +70,7 @@ const (
 
 func lsWorkspacesCommand() *cobra.Command {
 	var (
+		all       bool
 		outputFmt string
 		user      string
 		provider  string
@@ -85,11 +86,18 @@ func lsWorkspacesCommand() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			workspaces, err := getWorkspaces(ctx, client, user)
-			if err != nil {
-				return err
+			var workspaces []coder.Workspace
+			if !all {
+				var err error
+				workspaces, err = getWorkspaces(ctx, client, user)
+				if err != nil {
+					return err
+				}
+			} else {
+				// If the user gave the all flag, then filtering by user doesn't make sense.
+				user = ""
 			}
-			if provider != "" {
+			if provider != "" || all {
 				workspaces, err = getWorkspacesByProvider(ctx, client, provider, user)
 				if err != nil {
 					return err
@@ -124,6 +132,7 @@ func lsWorkspacesCommand() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().BoolVar(&all, "all", false, "Get workspaces for all users")
 	cmd.Flags().StringVar(&user, "user", coder.Me, "Specify the user whose resources to target")
 	cmd.Flags().StringVarP(&outputFmt, "output", "o", humanOutput, "human | json")
 	cmd.Flags().StringVarP(&provider, "provider", "p", "", "Filter workspaces by a particular workspace provider name.")
--- a/internal/cmd/workspaces.go
+++ b/internal/cmd/workspaces.go
@@ -87,18 +87,21 @@ func lsWorkspacesCommand() *cobra.Command {
 				return err
 			}
 			var workspaces []coder.Workspace
-			if !all {
+			if all {
 				var err error
-				workspaces, err = getWorkspaces(ctx, client, user)
+				workspaces, err = getAllWorkspaces(ctx, client)
+				if err != nil {
+					return err
+				}
+			} else if provider != "" {
+				var err error
+				workspaces, err = getWorkspacesByProvider(ctx, client, provider, user)
 				if err != nil {
 					return err
 				}
 			} else {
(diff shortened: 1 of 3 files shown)


internal/cmd/workspaces_test.go
```diff
@@ -29,6 +29,34 @@ func Test_workspaces_ls(t *testing.T) {
...
+	skipIfNoAuth(t)
+	for _, test := range []struct {
+		name    string
+		command []string
+		assert  func(r result)
+	}{
+		{
+			name:    "simple list",
+			command: []string{"workspaces", "ls", "--all"},
+			assert:  func(r result) { r.success(t) },
```
goodspark, line 42: Sanity check: This isn't actually checking the output - just that the command exited with 0?

internal/cmd/workspaces.go
```diff
@@ -124,6 +132,7 @@ func lsWorkspacesCommand() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().BoolVar(&all, "all", false, "Get workspaces for all users")
```
deansheather, line 135:
```suggestion
	cmd.Flags().BoolVar(&all, "all", false, "Get workspaces for all users (admin only)")
```

Only admins can get all workspaces by workspace provider it seems

internal/cmd/workspaces.go
```diff
@@ -85,11 +86,18 @@ func lsWorkspacesCommand() *cobra.Command {
...
+				workspaces, err = getWorkspaces(ctx, client, user)
+				if err != nil {
+					return err
+				}
+			} else {
+				// If the user gave the all flag, then filtering by user doesn't make sense.
+				user = ""
 			}
-			if provider != "" {
+			if provider != "" || all {
```
deansheather, line 100: This means the `--all` flag only works with the `--provider` flag. I think that's OK for now, but it needs to be made obvious in the flag description (e.g. "requires --provider flag") and there should be a check at the top of the command to ensure `--provider` is set if `--all` is set.

```
$ ./coder envs ls --all
fatal: resource not found
```

internal/cmd/workspaces_test.go
```diff
@@ -29,6 +29,34 @@ func Test_workspaces_ls(t *testing.T) {
...
+func Test_workspaces_ls_all(t *testing.T) {
+	skipIfNoAuth(t)
+	for _, test := range []struct {
+		name    string
+		command []string
+		assert  func(r result)
+	}{
+		{
+			name:    "simple list",
+			command: []string{"workspaces", "ls", "--all"},
```
deansheather, line 41: This and the test case below should have `--provider built-in`

deansheather (changes requested):

internal/cmd/workspaces_test.go
deansheather: Essentially, yeah. This mini test framework "runs" the commands in-process and checks if the command fn returned an error.

deansheather:
This looks great, just needs a few changes and it'll be solid 😄

internal/cmd/workspaces.go
goodspark: Doesn't it mean it works _regardless_ of the provider flag? If I call coder with just `--all`, provider is "", and the logic becomes `if "" != "" || true` -> `if false || true` -> `if true`. This was my intent - as an admin, I want to see _all_ workspaces.

But looking at the actual logic/implementation for getWorkspacesByProvider, I think your point still stands. I can fix that.

internal/cmd/workspaces.go
```diff
@@ -85,15 +86,25 @@ func lsWorkspacesCommand() *cobra.Command {
...
 				if err != nil {
 					return err
 				}
+			} else {
+				var err error
+				workspaces, err = getWorkspaces(ctx, client, user)
+				if err != nil {
+					return err
+				}
 			}
```
goodspark, line 108: Shuffled this around because it seemed like the first fetch for the user-only workspaces was ignored if providers was also passed.

internal/cmd/workspaces.go
```diff
@@ -85,15 +86,25 @@ func lsWorkspacesCommand() *cobra.Command {
...
 				return err
 			}
-			workspaces, err := getWorkspaces(ctx, client, user)
-			if err != nil {
-				return err
-			}
-			if provider != "" {
+			var workspaces []coder.Workspace
+			if all {
+				var err error
```
deansheather, line 91: You shouldn't need the `var err`s here or in the other if branches since it's already defined above, then you can condense the `if err` check to be underneath the if branches, so each branch is just the `workspaces, err = getX()` statement.

edit: check my suggestion

internal/cmd/workspaces.go
deansheather, line 108:
```suggestion
				workspaces, err = getAllWorkspaces(ctx, client)
			} else if provider != "" {
				workspaces, err = getWorkspacesByProvider(ctx, client, provider, user)
			} else {
				workspaces, err = getWorkspaces(ctx, client, user)
			}
			if err != nil {
				return err
			}
```

deansheather (approved):
Seems to work well when I checked out and ran it against our deployment

PR #417, opened 2021-08-12
feat: Add update command to coder-cli

johnstcn:
This commit adds a new update subcommand that queries a Coder instance for its current version, fetches the corresponding version from GitHub releases if required, and updates the binary in-place.

Currently, we only check Major.Minor.Patch version, and not pre-release. (Edit: checking pre-release now)
This will obviously not work if the corresponding release of coder-cli is not yet available.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.14
 require (
 	cdr.dev/slog v1.4.1
 	cdr.dev/wsep v0.0.0-20200728013649-82316a09813f
+	github.com/blang/semver/v4 v4.0.0
 	github.com/briandowns/spinner v1.16.0
 	github.com/cli/safeexec v1.0.0
 	github.com/fatih/color v1.12.0
@@ -23,6 +24,7 @@ require (
 	github.com/pion/webrtc/v3 v3.0.32
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/rjeczalik/notify v0.9.2
+	github.com/spf13/afero v1.6.0
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/pion/datachannel v1.4.21
 	github.com/pion/dtls/v2 v2.0.9
-	github.com/pion/ice/v2 v2.1.10
+	github.com/pion/ice/v2 v2.1.12
 	github.com/pion/logging v0.2.2
 	github.com/pion/turn/v2 v2.0.5
 	github.com/pion/webrtc/v3 v3.0.32
@@ -28,7 +28,7 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
-	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
+	golang.org/x/net v0.0.0-20210805182204-aaa1db679c0d
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20210615035016-665e8c7367d1
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	cdr.dev/slog v1.4.1
 	cdr.dev/wsep v0.0.0-20200728013649-82316a09813f
-	github.com/blang/semver/v4 v4.0.0
+	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/briandowns/spinner v1.16.0
 	github.com/cli/safeexec v1.0.0
 	github.com/fatih/color v1.12.0
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -130,7 +130,7 @@ func (u *updater) Run(ctx context.Context, force bool, coderURLString string) er
 		}
 	}
 
-	downloadURL := makeDownloadURL(desiredVersion.FinalizeVersion(), runtime.GOOS, runtime.GOARCH)
+	downloadURL := makeDownloadURL(desiredVersion, u.osF(), runtime.GOARCH)
 
 	var downloadBuf bytes.Buffer
 	memWriter := bufio.NewWriter(&downloadBuf)
@@ -197,14 +197,14 @@ func defaultConfirm(label string) (string, error) {
 	return p.Run()
 }
 
-func makeDownloadURL(version, ostype, archetype string) string {
+func makeDownloadURL(version *semver.Version, ostype, archtype string) string {
 	const template = "https://github.com/cdr/coder-cli/releases/download/v%s/coder-cli-%s-%s.%s"
 	var ext string
 	switch ostype {
 	case "linux":
 		ext = "tar.gz"
 	default:
-		ext = ".zip"
+		ext = "zip"
 	}
 	return fmt.Sprintf(template, version, ostype, archetype, ext)
 }
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -7,8 +7,10 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -307,21 +309,29 @@ func getCoderConfigURL() (*url.URL, error) {
 // XXX: coder.Client requires an API key, but we may not be logged into the coder instance for which we
 // want to determine the version. We don't need an API key to sniff the version header.
 func getAPIVersionUnauthed(client getter, baseURL url.URL) (*semver.Version, error) {
-	baseURL.Path = path.Join(baseURL.Path, "/api")
+	baseURL.Path = path.Join(baseURL.Path, "/api/private/version")
 	resp, err := client.Get(baseURL.String())
 	if err != nil {
 		return nil, xerrors.Errorf("get %s: %w", baseURL.String(), err)
 	}
 	defer resp.Body.Close()
 
-	versionHdr := resp.Header.Get("coder-version")
-	if versionHdr == "" {
-		return nil, xerrors.Errorf("URL %s response missing coder-version header", baseURL.String())
+	ver := struct {
+		Version string `json:"version"`
+	}{}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, xerrors.Errorf("read response body: %w", err)
 	}
 
-	version, err := semver.StrictNewVersion(versionHdr)
+	if err := json.Unmarshal(body, &ver); err != nil {
+		return nil, xerrors.Errorf("parse version response: %w", err)
+	}
+
+	version, err := semver.StrictNewVersion(ver.Version)
 	if err != nil {
-		return nil, xerrors.Errorf("parsing coder-version header: %w", err)
+		return nil, xerrors.Errorf("parsing coder version: %w", err)
 	}
 
 	return version, nil
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -56,9 +56,14 @@ func updateCmd() *cobra.Command {
 				Timeout: 10 * time.Second,
 			}
 
+			currExe, err := os.Executable()
+			if err != nil {
+				return clog.Fatal("init: get current executable", clog.Causef(err.Error()))
+			}
+
 			updater := &updater{
 				confirmF:       defaultConfirm,
-				executablePath: os.Args[0],
+				executablePath: currExe,
 				httpClient:     httpClient,
 				fs:             afero.NewOsFs(),
 				osF:            func() string { return runtime.GOOS },
@@ -85,7 +90,7 @@ func (u *updater) Run(ctx context.Context, force bool, coderURLString string) er
 
 	currentBinaryStat, err := u.fs.Stat(u.executablePath)
 	if err != nil {
-		return clog.Fatal("preflight: cannot stat current binary", clog.Causef("%s", err))
+		return clog.Fatal("preflight: cannot stat current binary", clog.Causef(err.Error()))
 	}
 
 	if currentBinaryStat.Mode().Perm()&0222 == 0 {
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -22,7 +22,7 @@ import (
 	"cdr.dev/coder-cli/pkg/clog"
 	"golang.org/x/xerrors"
 
-	"github.com/blang/semver/v4"
+	"github.com/Masterminds/semver/v3"
 	"github.com/manifoldco/promptui"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -116,7 +116,7 @@ func (u *updater) Run(ctx context.Context, force bool, coderURLString string) er
 	clog.LogInfo(fmt.Sprintf("Coder instance at %q reports version %s", coderURL.String(), desiredVersion.String()))
 	clog.LogInfo(fmt.Sprintf("Current version of coder-cli is %s", version.Version))
 
-	if currentVersion, err := semver.Make(u.versionF()); err == nil {
+	if currentVersion, err := semver.StrictNewVersion(u.versionF()); err == nil {
 		if desiredVersion.Compare(currentVersion) == 0 {
 			clog.LogInfo("Up to date!")
 			return nil
@@ -124,7 +124,7 @@ func (u *updater) Run(ctx context.Context, force bool, coderURLString string) er
 	}
 
 	if !force {
-		label := fmt.Sprintf("Update coder-cli to version %s", desiredVersion.FinalizeVersion())
+		label := fmt.Sprintf("Do you want to download version %s instead", desiredVersion)
 		if _, err := u.confirmF(label); err != nil {
 			return clog.Fatal("failed to confirm update", clog.Tipf(`use "--force" to update without confirmation`))
 		}
@@ -188,7 +188,7 @@ func (u *updater) Run(ctx context.Context, force bool, coderURLString string) er
 		return clog.Fatal("failed to update coder binary in-place", clog.Causef(err.Error()))
 	}
 
-	clog.LogSuccess("Updated coder CLI to version " + desiredVersion.FinalizeVersion())
+	clog.LogSuccess("Updated coder CLI to version " + desiredVersion.String())
 	return nil
 }
 
@@ -206,7 +206,18 @@ func makeDownloadURL(version *semver.Version, ostype, archtype string) string {
 	default:
 		ext = "zip"
 	}
-	return fmt.Sprintf(template, version, ostype, archetype, ext)
+	var b bytes.Buffer
+	fmt.Fprintf(&b, "%d", version.Major())
+	fmt.Fprint(&b, ".")
+	fmt.Fprintf(&b, "%d", version.Minor())
+	fmt.Fprint(&b, ".")
+	fmt.Fprintf(&b, "%d", version.Patch())
+	if version.Prerelease() != "" {
+		fmt.Fprint(&b, "-")
+		fmt.Fprint(&b, version.Prerelease())
+	}
+
+	return fmt.Sprintf(template, b.String(), ostype, archtype, ext)
 }
 
 func extractFromArchive(path string, archive []byte) ([]byte, error) {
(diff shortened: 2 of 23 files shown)


internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+			httpClient := &http.Client{
+				Timeout: 10 * time.Second,
+			}
+
+			updater := &updater{
+				httpClient:     httpClient,
+				fs:             afero.NewOsFs(),
+				confirmF:       defaultConfirm,
+				versionF:       func() string { return version.Version },
+				executablePath: os.Args[0],
```
deansheather, line 61: nit: use the same order as the struct

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return configuredURL, nil
+}
+
+// XXX: coder.Client requires an API key, but we may not be logged into the coder instance for which we
+// want to determine the version. We don't need an API key to sniff the version header.
+func getAPIVersionUnauthed(client getter, baseURL url.URL) (semver.Version, error) {
+	baseURL.Path = path.Join(baseURL.Path, "/api")
```
deansheather, line 291: Use `/api/private/version` over the header. It's a public route.

internal/cmd/update_test.go
```diff
@@ -0,0 +1,215 @@
...
+	fakeExePath    = "/coder"
+	fakeCoderURL   = "https://my.cdr.dev"
+	fakeNewVersion = "1.23.4-rc.5+678-gabcdef-12345678"
+	fakeOldVersion = "1.22.4-rc.5+678-gabcdef-12345678"
+	fakeReleaseURL = "https://github.com/cdr/coder-cli/releases/download/v1.23.4/coder-cli-linux-amd64.tar.gz"
+)
+
+func Test_updater_run_noop(t *testing.T) {
+	fakefs := afero.NewMemMapFs()
+	httpClient := &fakeGetter{
```
deansheather, line 29: Lots of duped code in these tests, consider making a `run()`-like function like we have in the monorepo

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+
+	_ = resp.Body.Close()
+
+	if err := memWriter.Flush(); err != nil {
+		return clog.Fatal(fmt.Sprintf("failed to save %s", downloadURL), clog.Causef(err.Error()))
+	}
+
+	// TODO: validate the checksum of the downloaded file. GitHub does not currently provide this information
+	// and we do not generate them yet.
+	updatedBinary, err := extractFromArchive("coder", downloadBuf.Bytes())
```
deansheather, line 158: `coder.exe` on windows

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+type getter interface {
+	Get(url string) (*http.Response, error)
+}
+
+func (u *updater) Run(ctx context.Context, force bool, coderURLString string) error {
+	// TODO: check under following directories and warn if coder binary is under them:
+	//   * homebrew prefix
+	//   * coder assets root (env CODER_ASSETS_ROOT)
```
kylecarbs, line 80: Ahh wise check!

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+	return cmd
+}
+
+type getter interface {
+	Get(url string) (*http.Response, error)
+}
+
+func (u *updater) Run(ctx context.Context, force bool, coderURLString string) error {
+	// TODO: check under following directories and warn if coder binary is under them:
+	//   * homebrew prefix
```
kylecarbs, line 79: Should we run `brew update ...` for them if that's the case?

johnstcn: I'd find that a bit presumptuous myself. It's probably safer to tell users to run brew update manually.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.14
 require (
 	cdr.dev/slog v1.4.1
 	cdr.dev/wsep v0.0.0-20200728013649-82316a09813f
+	github.com/blang/semver/v4 v4.0.0
```
jawnsy, line 8: How does this library compare to https://github.com/Masterminds/semver? That one is used in Helm, see https://github.com/helm/helm/blob/efe2638f87b597403f34009e1029a6f0b44db8f2/go.mod#L8

internal/cmd/update.go
kylecarbs: Good point!

go.mod
jawnsy: They both seem to have roughly the same number of stars and dependent libraries :raised_eyebrow: 

https://github.com/blang/semver/network/dependents?dependent_type=REPOSITORY&package_id=UGFja2FnZS0yMjY0ODE0Njcy

https://github.com/Masterminds/semver/network/dependents?package_id=UGFja2FnZS0yMjY0ODEyOTkw

I'm asking because the coder-doctor should use the same library, whichever it is. I'm not familiar with the blang one, but am open to switching if it's better!

go.mod
```diff
@@ -23,6 +24,7 @@ require (
 	github.com/pion/webrtc/v3 v3.0.32
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/rjeczalik/notify v0.9.2
+	github.com/spf13/afero v1.6.0
```
jawnsy, line 27: Is this dependency still necessary, now that Go has a built-in in-memory FS for testing? https://pkg.go.dev/testing/fstest

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+	desiredVersion, err := getAPIVersionUnauthed(u.httpClient, *coderURL)
+	if err != nil {
+		return clog.Fatal("fetch api version", clog.Causef(err.Error()))
+	}
+
+	clog.LogInfo(fmt.Sprintf("Coder instance at %q reports version %s", coderURL.String(), desiredVersion.String()))
+	clog.LogInfo(fmt.Sprintf("Current version of coder-cli is %s", version.Version))
+
+	if currentVersion, err := semver.Make(u.versionF()); err == nil {
+		if desiredVersion.Compare(currentVersion) == 0 {
```
jawnsy, line 118: Won't this check also fail if the client is newer than the server? I think that's the correct behavior, but it's not necessarily that the client is newer or up to date with the server, what we are really checking is that the versions match, not necessarily that anything is up-to-date

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// TODO: validate the checksum of the downloaded file. GitHub does not currently provide this information
+	// and we do not generate them yet.
+	updatedBinary, err := extractFromArchive("coder", downloadBuf.Bytes())
+	if err != nil {
+		return clog.Fatal("failed to extract coder binary from archive", clog.Causef(err.Error()))
+	}
+
+	// We assume the binary is named coder and write it to coder.new
+	updatedCoderBinaryPath := u.executablePath + ".new"
```
jawnsy, line 164: An implicit assumption of this code is that users either:

1. Only use a single coder installation; or
2. Use the same version of coder across multiple deployments (e.g. dev/stage/test)

I don't think we should make this assumption in general, and would propose a scheme where coder-cli has a bootstrap portion which looks for existing commands, updates if needed, and then exec's that. For example:

* `coder` command is a small shim that does this release check
* check the server version (suppose for this example it's 1.20.0)
* if the versions don't match, download the corresponding version, and save as `~/.coder/bin/coder-1.20.0.exe`)
* exec that command and pass along arguments

This way, a user can transparently switch between different installations of Coder, and always have an up-to-date version of the tool. And we don't have to worry about long-term compatibility guarantees between client and server (for now)

go.mod
johnstcn: Unfortunately io/fs only supports read-only right now. But there's a [proposal in the works](https://github.com/golang/go/issues/45757) to support a writeable filesystem interface!

internal/cmd/update.go
johnstcn: Yes, so it's probably a case of rewording some messages emitted by the tool - "Update to version X-1" sounds very strange.

internal/cmd/update.go
jawnsy: One other thing to consider -- the `oc` command (for OpenShift) and busybox have magic behavior where the command that runs depends on `argv[0]`, which means that we could reuse the `coder` CLI command for the bootstrap logic as well as the command itself. That is, if the program is run as `coder`, then do the bootstrapping I described above, and if the program is run as `coder-1.21.0`, then run the coder CLI commands.

This means that we don't have to create a new build or change the way things are built or delivered, but we benefit from the above logic

For the way oc does it, check here: https://github.com/openshift/oc/blob/25b5251c48ea8e3d1294e6b9b1c05bebb29de46f/cmd/oc/oc.go#L91-L93

go.mod
Emyrk: We use Masterminds in `/m` too: https://github.com/cdr/m/blob/master/go.mod#L18

Emyrk: And we use: `https://github.com/hashicorp/go-version`

johnstcn: I tried out Masterminds and it seems more feature-complete. Switching to that.

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+
+func (u *updater) Run(ctx context.Context, force bool, coderURLString string) error {
+	// TODO: check under following directories and warn if coder binary is under them:
+	//   * homebrew prefix
+	//   * coder assets root (env CODER_ASSETS_ROOT)
+
+	currentBinaryStat, err := u.fs.Stat(u.executablePath)
+	if err != nil {
+		return clog.Fatal("preflight: cannot stat current binary", clog.Causef("%s", err))
+	}
```
Emyrk, line 85: won't this fail if `coder` is in my `PATH`? If we cannot handle this case, maybe we should report some nicer error or allow using a flag to pass the path.

`which coder` can find it, but idk if that works on windows...

internal/cmd/update.go
```diff
@@ -0,0 +1,309 @@
...
+		return clog.Fatal("failed to write updated coder binary to disk", clog.Causef(err.Error()))
+	}
+
+	if err := fsWriter.Flush(); err != nil {
+		return clog.Fatal("failed to persist updated coder binary to disk", clog.Causef(err.Error()))
+	}
+
+	if err = u.fs.Rename(updatedCoderBinaryPath, u.executablePath); err != nil {
+		return clog.Fatal("failed to update coder binary in-place", clog.Causef(err.Error()))
+	}
```
Emyrk, line 181: Can we test that we can execute the new binary before we rename?

internal/cmd/update.go
```diff
@@ -0,0 +1,395 @@
...
+
+	if !force {
+		label := fmt.Sprintf("Do you want to download version %s instead", desiredVersion)
+		if _, err := u.confirmF(label); err != nil {
+			return clog.Fatal("failed to confirm update", clog.Tipf(`use "--force" to update without confirmation`))
+		}
+	}
+
+	downloadURL := makeDownloadURL(desiredVersion, u.osF(), runtime.GOARCH)
```
kylecarbs, line 162: Adding a `--tag` flag to explicitly specify a version to download could be helpful. I could imagine us publishing an RC and wanting to `coder update --tag` to that. Another option would be `coder update --next` to use prereleases.

internal/cmd/update.go
johnstcn: I tested this on Windows and it seems to work fine.

internal/cmd/update.go
johnstcn: Having `coder-cli` be able to manage multiple versions of itself would be a major usability win.

However, I think it's a bit out of scope of this issue, so I created a separate story to follow up on it.

internal/cmd/update.go
johnstcn: Done

internal/cmd/update.go
johnstcn: Done, but `coder update --version <version_spec>` instead.

kylecarbs (approved):
Tested locally and works wonderfully!

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+)
+
+// updater updates coder-cli.
+type updater struct {
+	confirmF       func(string) (string, error)
+	execF          func(context.Context, string, ...string) ([]byte, error)
+	executablePath string
+	fs             afero.Fs
+	httpClient     getter
+	osF            func() string
```
deansheather, line 48: osF can just be a string instead of a function.

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+	cmd.Flags().StringVar(&versionArg, "version", "", "explicitly specify which version to fetch and install")
+
+	return cmd
+}
+
+type getter interface {
+	Get(url string) (*http.Response, error)
+}
+
+func (u *updater) Run(ctx context.Context, force bool, coderURLArg string, versionArg string) error {
```
deansheather, line 98: This function could benefit from being split up a little bit more, it's fairly long as it stands right now

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	// TODO: validate the checksum of the downloaded file. GitHub does not currently provide this information
+	// and we do not generate them yet.
+	var updatedBinaryName string
+	if u.osF() == "windows" {
+		updatedBinaryName = "coder.exe"
+	} else {
+		updatedBinaryName = "coder"
+	}
```
deansheather, line 187:
```suggestion
	updatedBinaryName := "coder"
	if u.osF() == "windows" {
		updatedBinaryName = "coder.exe"
	}
```

internal/cmd/update_test.go
```diff
@@ -0,0 +1,542 @@
...
+			OsF: func() string { return goosLinux },
+			// This must be overridden inside run()
+			VersionF: func() string {
+				t.Errorf("unhandled VersionF")
+				t.FailNow()
+				return ""
+			},
+		}
+
+		fn(t, params)
```
deansheather, line 101:
```suggestion
		t.Run(name, func(t *testing.T) {
			ctx := context.Background()
			fakefs := afero.NewMemMapFs()
			execer := newFakeExecer(t)
			execer.M["brew --prefix"] = fakeExecerResult{[]byte{}, os.ErrNotExist}
			params := &params{
				// This must be overridden inside run()
				ConfirmF: func(string) (string, error) {
					t.Errorf("unhandled ConfirmF")
					t.FailNow()
					return "", nil
				},
				Execer:         execer,
				Ctx:            ctx,
				ExecutablePath: fakeExePathLinux,
				Fakefs:         fakefs,
				HTTPClient:     newFakeGetter(t),
				// Default to GOOS=linux
				OsF: func() string { return goosLinux },
				// This must be overridden inside run()
				VersionF: func() string {
					t.Errorf("unhandled VersionF")
					t.FailNow()
					return ""
				},
			}

			fn(t, params)
		})
```

and possibly put `t.Parallel()` first inside t.Run if these tests can be run in parallel

deansheather: this diff is awful, tldr I wrapped everything in t.Run

internal/cmd/update_test.go
```diff
@@ -0,0 +1,542 @@
...
+	apiPrivateVersionURL  = fakeCoderURL + apiPrivateVersion
+	fakeError             = xerrors.New("fake error for testing")
+	fakeNewVersionJSON    = fmt.Sprintf(`{"version":%q}`, fakeNewVersion)
+	fakeOldVersionJSON    = fmt.Sprintf(`{"version":%q}`, fakeOldVersion)
+	fakeAssetURLLinux     = "https://github.com/cdr/coder-cli/releases/download/v1.23.4-rc.5/" + filenameLinux
+	fakeAssetURLWindows   = "https://github.com/cdr/coder-cli/releases/download/v1.23.4-rc.5/" + filenameWindows
+	fakeGithubReleaseJSON = fmt.Sprintf(`{"assets":[{"name":%q,"browser_download_url":%q},{"name":%q,"browser_download_url":%q}]}`, filenameLinux, fakeAssetURLLinux, filenameWindows, fakeAssetURLWindows)
+)
+
+func Test_updater_run(t *testing.T) {
```
deansheather, line 45: I love how the tests cover other platforms too, very nice.

deansheather (approved):

deansheather:
Also I think for the initial version we shouldn't worry about querying github API, the release artifacts are consistent right now. This PR is getting pretty big

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+		`C:\Windows\`,
+		`/var/tmp/coder`,
+	}
+	brewPrefixCmd, err := u.execF(ctx, "brew", "--prefix")
+	if err == nil { // ignore errors if homebrew not installed
+		pathBlockList = append(pathBlockList, strings.TrimSpace(string(brewPrefixCmd)))
+	}
+
+	for _, prefix := range pathBlockList {
+		if HasFilePathPrefix(u.executablePath, prefix) {
```
jawnsy, line 113: This behavior is good for now, just wanted to leave a suggestion for your consideration in a future iteration: we could also write the file to a well-known location (e.g. `~/.local/coder/bin`) or similar, and then use LookPath to see if we find the file we just wrote. If the version doesn't match, then we know that the user has something else earlier on their path obscuring the location of where we wrote the file, so we could issue a message in that case instructing users what to do (e.g. move ~/.local/coder/bin earlier in the PATH, or remove the other one from the PATH, etc)

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+	currentVersion, err := semver.StrictNewVersion(u.versionF())
+	if err != nil {
+		clog.LogWarn("failed to determine current version of coder-cli", clog.Causef(err.Error()))
+	} else if currentVersion.Compare(desiredVersion) == 0 {
+		clog.LogInfo("Up to date!")
+		return nil
+	}
+
+	if !force {
+		label := fmt.Sprintf("Do you want to download version %s instead", desiredVersion)
```
jawnsy, line 146: should the prompt have a `?` and `[Y/n]` to show the default and available options? Something like:

`Do you want to download version 1.21.0 instead? [Y/n]`

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+		if _, err := u.confirmF(label); err != nil {
+			return clog.Fatal("failed to confirm update", clog.Tipf(`use "--force" to update without confirmation`))
```
jawnsy, line 148: instead of "failed to confirm update", I think "user cancelled operation" or similar would be clearer?

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, xerrors.Errorf("failed to read tar archive: %w", err)
+		}
+		fi := hdr.FileInfo()
+		if fi.Name() == path && fi.Mode().IsRegular() {
+			_, _ = io.Copy(bw, tr)
```
jawnsy, line 386: seems like we should handle this error instead of ignoring it? I guess it could be that the tar we downloaded is corrupt?

johnstcn:
> Also I think for the initial version we shouldn't worry about querying github API, the release artifacts are consistent right now. This PR is getting pretty big

I found the opposite, we have `coder-cli-linux-amd64.zip` vs `coder-cli-windows.zip` or `coder-cli-windows-386.zip` and so on. Querying the GitHub releases API seems less failure-prone here.

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("get %s: %w", baseURL.String(), err)
+	}
+	defer resp.Body.Close()
+
+	ver := struct {
+		Version string `json:"version"`
+	}{}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
jawnsy, line 421: Ah, we should consider increasing our version requirement for coder-cli to 1.16 so that we can use io.ReadAll instead (ioutil stuff is now deprecated). No rush to do that, though, we can do it after 1.17 is released (should be soon)

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+		return nil, err
+	}
+	configuredURL, err := url.Parse(strings.TrimSpace(urlString))
+	if err != nil {
+		return nil, err
+	}
+	return configuredURL, nil
+}
+
+// XXX: coder.Client requires an API key, but we may not be logged into the coder instance for which we
```
jawnsy, line 407: is this still true with the unauthenticated API?

internal/cmd/update.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	return version, nil
+}
+
+// HasFilePathPrefix reports whether the filesystem path s
+// begins with the elements in prefix.
+// Lifted from github.com/golang/go/blob/master/src/cmd/internal/str/path.go.
+func HasFilePathPrefix(s, prefix string) bool {
+	sv := strings.ToUpper(filepath.VolumeName(s))
```
jawnsy, line 442: I'm always suspicious of ToUpper/ToLower because those often don't handle locales correctly. Would using functions like `strings.EqualFold` work here? https://pkg.go.dev/strings#EqualFold

jawnsy (approved):
LGTM, my comments are minor, feel free to address them in a follow-up (or not at all).

internal/cmd/update.go
johnstcn: Both of these are done automatically by `promptui`.

internal/cmd/update.go
johnstcn: I don't think there's any change here. There's no way I can see to instantiate a `coder.Client` without credentials. In any case, it looks like the client just sniffs the HTTP headers to fetch the version.

internal/cmd/update.go
johnstcn: It appears to handle this case fine:

```
p1 := `/Users/Bürgermeister/`
p2 := `/Users/Bürgermeister/Programmierung`
assert.True(t, "test HasFilePathPrefix", HasFilePathPrefix(p2, p1)) // passes
```

PR #424, opened 2021-09-02
feat: add pprof to agent

coadler:


internal/cmd/agent.go
```diff
@@ -78,6 +81,8 @@ coder agent start --coder-url https://my-coder.com --token xxxx-xxxx
 				}
 			}
 
+			go servePprof(ctx, log)
```
jawnsy, line 84: Hmm... since [coder cli has pprof](https://github.com/cdr/coder-cli/blob/59a0a209646960bfc3fbd7576a9513b9b76dfd8e/cmd/coder/main.go#L22-L26) already, is there some way to share that code regardless of whether someone is running a normal coder command or coder agent?

jawnsy (approved):

deansheather (commented):
Keep the existing pprof code, but force it on in `coder agent` command, perhaps by checking argv[1]

PR #434, opened 2021-09-09
chore: cleanup wsnet/rtc

coadler:
Makes code a bit easier to read and uses xerrors for all errors.


wsnet/rtc.go
```diff
@@ -220,65 +264,86 @@ func proxyICECandidates(conn *webrtc.PeerConnection, w io.Writer) func() {
...
-// Waits for a DataChannel to hit the open state.
+// waitForDataChannelOpen waits for a DataChannel to hit the open state.
 func waitForDataChannelOpen(ctx context.Context, channel *webrtc.DataChannel) error {
-	if channel.ReadyState() == webrtc.DataChannelStateOpen {
+	switch channel.ReadyState() {
+	case webrtc.DataChannelStateOpen:
 		return nil
+
+	case webrtc.DataChannelStateClosed,
+		webrtc.DataChannelStateClosing:
```
deansheather, line 319: I think we should revert this but maybe reword the error to "not connecting"

deansheather (approved):

PR #441, opened 2021-09-13
fixup! feat: Add update command to coder-cli (#417)

johnstcn:
* Use `semver.NewVersion` instead of `semver.StrictNewVersion` as otherwise users must type the explicit trailing zeros e.g. `coder update --version 1.21.0` instead of `coder update --version 1.21`.


jawnsy (approved):
consider adding tests! unfortunately there's no easy way to unit test this stuff though :/

johnstcn:
@jawnsy added some quick tests for that!

PR #444, opened 2021-09-14
feat: Allow using custom certs for `coder agent` to use

Emyrk:
# What this does

This allows using custom certs for `coder agent` to use when communicating with a coder deployment. I intentionally do not place the certs in [`/etc/ssl/certs`](https://golang.org/src/crypto/x509/root_linux.go) as I do not want to add certs for the whole workspace. I only want to add certs for the `agent <--> deployment` communication.

In Go 1.17 we can use https://pkg.go.dev/crypto/x509#SystemCertPool to expand the existing cert pool, but until then when adding custom certs, those are the ONLY certs that are trusted.


kylecarbs (changes requested):
In https://github.com/cdr/coder-cli/pull/440/files I abstract the `Listen` function to provide options. It could use similar code and inject an HTTP client instead.

We don't want to set the precedent of creating new `Listen` funcs for various transport funcs... so might be good to abstract that now.

Emyrk:
> In https://github.com/cdr/coder-cli/pull/440/files I abstract the `Listen` function to provide options. It could use similar code and inject an HTTP client instead.
> 
> We don't want to set the precedent of creating new `Listen` funcs for various transport funcs... so might be good to abstract that now.

Good idea, you actually have `HTTPClient` as an option it looks like: https://github.com/cdr/coder-cli/blob/93d624dcc34374c35e3c23b8b0dd324cccaddac4/wsnet/listen.go#L50

which accomplishes my goal. Can we pick out that refactor from your draft? The ListenOptions?

kylecarbs:
Yup. Feel free to pick it out,... I'll merge back into mine once this is in.

Emyrk:
No longer need this

PR #446, opened 2021-09-22
allow coder login for WSL

arminaaki:
## Issue
`coder login` fails to open user's browser while running from [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/).
```
$ coder login <coder_url>
/usr/bin/xdg-open: 869: www-browser: not found
/usr/bin/xdg-open: 869: links2: not found
/usr/bin/xdg-open: 869: elinks: not found
/usr/bin/xdg-open: 869: links: not found
/usr/bin/xdg-open: 869: lynx: not found
/usr/bin/xdg-open: 869: w3m: not found
```

## Changes
* allow `coder login` to work while running from WSL.
* Detect if running in WSL by identifying `microsoft` in  `/proc/version` file.
```
$ cat /proc/version
Linux version ....-Microsoft (Microsoft@Microsoft.com) (gcc version ... (GCC) ) #...-Microsoft ....
```

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -143,7 +143,7 @@ func openURL(url string) error {
 	if wsl {
 		cmd = "cmd.exe"
 		args = []string{"/c", "start"}
-		url = strings.Replace(url, "&", "^&", -1)
+		url = strings.ReplaceAll(url, "&", "^&")
 		args = append(args, url)
 		return exec.Command(cmd, args...).Start()
 	}
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -120,7 +120,7 @@ func pingAPI(ctx context.Context, workspaceURL *url.URL, token string) error {
 
 // isWSL determines if coder-cli is running within Windows Subsystem for Linux
 func isWSL() (bool, error) {
-	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+	if runtime.GOOS == goosDarwin || runtime.GOOS == goosWindows {
 		return false, nil
 	}
 	data, err := ioutil.ReadFile("/proc/version")
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -4,7 +4,10 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/url"
+	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/pkg/browser"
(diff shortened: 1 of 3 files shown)


vapurrmaid (commented):
Code looks great to me, appreciate the extra log! Thanks for the contribution! 👍🏻

internal/cmd/login.go
```diff
@@ -113,3 +117,36 @@ func pingAPI(ctx context.Context, workspaceURL *url.URL, token string) error {
 	}
 	return nil
 }
+
+// isWSL determines if coder-cli is running within Windows Subsystem for Linux
+func isWSL() (bool, error) {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
```
vapurrmaid, line 123: it looks like we should make use of constants that are already defined:

https://github.com/cdr/coder-cli/blob/582f213d183be7135e2fcf0e6f839ea320e6c2f3/internal/cmd/update.go#L37

we should add `goosDarwin` as a const as well

internal/cmd/login.go
```diff
@@ -113,3 +117,36 @@ func pingAPI(ctx context.Context, workspaceURL *url.URL, token string) error {
...
+
+	wsl, err := isWSL()
+	if err != nil {
+		return xerrors.Errorf("test running Windows Subsystem for Linux: %w", err)
+	}
+
+	if wsl {
+		cmd = "cmd.exe"
+		args = []string{"/c", "start"}
+		url = strings.Replace(url, "&", "^&", -1)
```
vapurrmaid, line 146:
```suggestion
		url = strings.ReplaceAll(url, "&", "^&")
```

internal/cmd/login.go
```diff
@@ -113,3 +117,36 @@ func pingAPI(ctx context.Context, workspaceURL *url.URL, token string) error {
...
+	}
+	return strings.Contains(strings.ToLower(string(data)), "microsoft"), nil
+}
+
+// openURL opens the provided URL via user's default browser
+func openURL(url string) error {
+	var cmd string
+	var args []string
+
+	wsl, err := isWSL()
```
jawnsy, line 138: I think this is good, but I do wonder whether we could propose this upstream to pkg/browser

arminaaki: Thank you for the review. I agree, this work should ideally be  part of upstream `pkg/browser`. But Thoughts on moving forward with the work for Coder until I have the patch ready for upstream?

internal/cmd/login.go
arminaaki: Thank you for your review. Updated!

internal/cmd/login.go
arminaaki: Done!

internal/cmd/login.go
arminaaki: This may resolve  → https://github.com/cdr/coder-cli/pull/448

internal/cmd/login.go
jawnsy: Yeah, I think it makes sense to make this change now for sure! No objection to it :)

Thanks for your contribution!

vapurrmaid (approved):

PR #448, opened 2021-09-24
refactor: use constants for OSs

greyscaled:
Inspired from https://github.com/cdr/coder-cli/pull/446/checks?check_run_id=3695106435

### DIFF
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -36,6 +36,7 @@ import (
 const (
 	goosWindows       = "windows"
 	goosLinux         = "linux"
+	goosDarwin        = "darwin"
 	apiPrivateVersion = "/api/private/version"
 )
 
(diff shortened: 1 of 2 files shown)


f0ssel (approved):

internal/cmd/update.go
```diff
@@ -36,6 +36,7 @@ import (
 const (
 	goosWindows       = "windows"
```
jawnsy, line 37: It's unfortunate that Go doesn't seem to offer these constants in their code anywhere. It looks like Go themselves use hardcoded string comparisons everywhere: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/golang/go%24+%22windows%22&patternType=literal -- given that, maybe it's okay for us to do that too?

jawnsy (approved):

internal/cmd/update.go
vapurrmaid: Personally I think that doesn't sit well unless at least a type (like TS union of strings as a type alias) is defined. That feels a bit like walking on quicksand.

I don't know their reason, but I think it's reasonable we reach for them in our own domain model/application code base. I'd expect there to be some concept of a constant, enumeration or type in other languages when approaching a similar problem. Go sometimes makes odd choices IMO :P

